PR #1, opened 2025-09-30
feat(mqtt): add MQTT client interface with Paho as an default implementation

sauk2:
## Summary

This PR introduces an MQTT client interface with Paho as the default implementation.
It also adds a base logger that is decoupled from any specific logging library.

Unit and integration tests are included, with example usage demonstrated in the integration tests.


Briancbn (changes requested):
Thanks @sauk2 good work!

I have a few suggestion
- Use `<snake_case>.cpp` to be consistent with other RMF2 packages
- Move the `PahoMqttClient.hpp` to the `include`, let's build APIs with the intention that they can be tested and used by other packages.
- For the Logger, I have [a similar implementation](https://github.com/ros-industrial/rmf_scheduler/blob/develop/v1/rmf2_scheduler/include/rmf2_scheduler/log.hpp). You can reference, it might not be better. This is [the example I referenced](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/master/include/ur_client_library/log.h).

Briancbn (approved):
Good work!

PR #3, opened 2025-10-13
feat(order_execution): adds OrderManager

shawnkchan:
## Summary

Raising this PR to get feedback on my implementation for the order execution logic. This PR covers Order execution - Order Parsing and Order execution - Node Sequencing.

This PR contains the following features:
1. OrderManager class
    * Receives an Order message from the Order topic
    * Validates the order
    * Exposes API for OrderExecutor to sequence through the nodes and edges of the Order
2. GraphValidator class
    * Checks that the graph in an Order is valid
3. OrderGraphElement class
    *  Parent class to Node and Edge class
4. Node class
5. Edge class
6. Order Class
7. Unit tests for OrderManager and GraphValidator

NOTE: `IStateManager.hpp` is a temporary class for mocking the StateManger in the unit tests.


sauk2:
Thanks for all the great work on this!
This is a solid iteration and the direction looks good. To make it easier to review, test and merge this safely, I would like to suggest breaking the work into smaller pieces. This will make it easier to validate each part and keep the `main` stable.

Some high-level feedback:
- Retarget this PR to `develop` instead as we are holding off on pushing larger changes directly to `main`.
- Let’s merge PR #6 first as it introduces hard types that both the `OrderManager` and `StateManager` can use. Getting those in first will reduce a lot of duplication.
- Split the order and state logic into separate branches. Right now, this PR mixes order-related code with state-related code. To keep responsibilities clear, 
  - `OrderManager` should focus strictly on order handling.
  - `StateManager` should focus strictly on state tracking.
  - The execution layer should serve as the bridge between the two.
  
  This separation would help us avoid having order-related classes directly mutate state and it keeps the architecture more testable.

Briancbn:
Thanks @shawnkchan for this initial prototype, we will continue this work in the using the `vda5050_core::execution`. I will proceed to close this PR.

PR #5, opened 2025-10-20
Fix on Issue #4: Required APIs from StateMananger.

### LINKED ISSUE #4: Required APIs from StateMananger.
@Johnarman1398 Here's the list of APIs that my OrderManager will need from StateManager. Feel free to rename the functions to whatever you think is appropriate.

1. `last_node_id()`:
    * return nodeId of the latest node that the AGV has reached
2. `last_node_sequence_id()`:
    * sequenceId of the latest node that the AGV has reached
3. `is_node_states_empty` 
    * returns `true` if the `nodeStates` array maintained by `StateManager` contains zero nodes, `false` otherwise
4. `are_action_states_still_executing()` 
    * returns `true` if the `actionStates` array maintained by `StateManager` has actions that are not FINISHED or FAILED, `false` otherwise
6. `cleanup_previous_order()`
    * Clears any member variables to do with an existing order stored by the StateManager
    * This function will be called before the OrderManager sets a new order on the AGV
7. `set_new_order(const vda5050_core::order::Order& order)`
    * function to set a new order on the vehicle
    * for when vehicle accepts a new order
    * occurs after vehicle clears existing order
8. `clear_horizon()`
    * function to clear the horizon nodes or edges from the vehicle’s `edgeStates` array and `nodeStates` array
    * occurs when an order update is sent that updates a partially released order OR that updates an order that is still being executed
9. `append_states_for_update(vda5050_core::order::Order& order_update)`
    * function to append order update to the vehicle’s current order
    * occurs when the vehicle needs to modify its current `nodeStates` and `edgeStates` to add the order update

Let me know if anything isn't clear, thanks!

Johnarman1398:
To merge status manager and order manager, Fix #4


vda5050_core/src/vda5050_core/state_manager/state_manager.cpp
```diff
@@ -0,0 +1,380 @@
...
+  auto node_predicate = [](const NodeState& n) { return !n.released; };
+  auto edge_predicate = [](const EdgeState& e) { return !e.released; };
+
+  nodes.erase(
+    std::remove_if(nodes.begin(), nodes.end(), node_predicate), nodes.end());
+  edges.erase(
+    std::remove_if(edges.begin(), edges.end(), edge_predicate), edges.end());
+}
+
+// TODO @(johnaa) sounds like it does the same feature as set_new_order, to clear with @shawnkchan
```
shawnkchan, line 367: Good point! I see why these two could look the same. 

I think this boils down to whether we are continuing to store the `nodeStates` and `edgeStates` of a completed order on the AGV. My thinking is that once we know the AGV has complete its currently stored order, and a new order arrives, we would not need any information from the completed order. We can delete the completed order and just add the new order to the AGV's `StateManager`, saving storage on the AGV. If I understand correctly, `set_new_order` 's current implementation continues to store the previous `nodeStates` and `edgeStates`. 

Here's a detailed description of how I envisioned the functions:
`set_new_order` gets called when the AGV no longer has any actions to carry out. This is as the AGV has completed the order that its `StateManager` currently holds, and hence it is safe to delete all existing `nodeStates` and `edgeStates` I.e: all actions from the edges and nodes of the existing order have been executed. 

`append_states_for_update` gets called when the incoming order is an update of the existing order. In this scenario, the AGV could still have actions that it has not yet executed from its existing order. Hence, we want `StateManager` to continue to store the existing order and append the incoming order to the existing order. 

Let me know if anything I explained is unclear! I understand that it's pretty verbose.

vda5050_core/src/vda5050_core/state_manager/state_manager.cpp
```diff
@@ -0,0 +1,380 @@
...
+      action_state.action_status != ActionStatus::FINISHED &&
+      action_state.action_status != ActionStatus::FAILED)
+    {
+      return true;
+    }
+  }
+  return false;
+}
+
+void StateManager::cleanup_previous_order()
```
shawnkchan, line 274: I envisioned that this function handles the 'delete state state of previous order' box in the flowchart on page 18 of the VDA5050 2.1.0 Recommendation (I refer to the 2.1.0 flowchart as it describes the same order handling logic as the 2.0.0 version, but much easier to read). 

Hence, this function would need to carry out these effects too:
1. Update `robot_state_.last_node_sequence_Id`
2. Clear `nodeStates` array
3. Clear `edgeStates` array

Effects 2 and 3 are in preparation for `set_new_order`. My comment on line 367 explains why this is needed.

> I think this boils down to whether we are continuing to store the nodeStates and edgeStates of a completed order on the AGV. My thinking is that once we know the AGV has complete its currently stored order, and a new order arrives, we would not need any information from the completed order. We can delete the completed order and just add the new order to the AGV's StateManager, saving storage on the AGV. If I understand correctly, set_new_order 's current implementation continues to store the previous nodeStates and edgeStates.

shawnkchan (changes requested):
Thanks for adding the features John! I've added some comments that clarify your question on the `set_new_order` and `append_states_for_update` functions, as well as a remark regarding `cleanup_previous_order`.

Issue #8, opened 2025-11-18
Extract order and graph validation related logic and target to develop


sauk2:
## Summary

We need to isolate the order and graph validation related code in #3 so they can be developed and tested independently of `OrderManager` and `StateManager`. This will form the base for subsequent PRs related to order. The general idea is to,

- [ ] Move all the order validation related code into a separate branch (`OrderGraphValidator` and associated classes)
- [ ] Use the hard types introduced as part of PR #6 
- [ ] Make sure there is no interaction with the state logic
- [ ] Write unit tests

PR #13, opened 2025-11-24
refactor(orderGraphValidator): Isolate OrderGraphValidator and use `vda5050_types` 

shawnkchan:
## Summary
This PR addresses https://github.com/ros-industrial/vda5050_client/issues/8. This branch was created from the `develop` branch.

In this branch:
1. OrderGraphValidator is now isolated from OrderManager
2. OrderGraphValidator uses the hard types in `vda5050_types`
3. There is no interaction with any state logic
4. Unit tests are passing


vda5050_core/CMakeLists.txt
```diff
@@ -33,11 +34,23 @@ target_include_directories(logger
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
 
+add_library(order_graph_validator src/vda5050_core/order_graph_validator/order_graph_validator.cpp)
```
sauk2: Let's package all the order and state related code into a `client` library. The following `OrderManager` PR will go into this library too

vda5050_core/include/vda5050_core/client/order/order_graph_validator.hpp
sauk2, line 1: We probably don't need a separate `order_graph_validator/` folder. I was thinking more like `include/vda5050_core/client/order/...` to hold all code associated to order on the client side. We will be adding in the master related code too so that will be in the folder `include/vda5050_core/master/...`. Same for `src/...`

vda5050_core/include/vda5050_core/order_graph_validator/order_graph_validator.hpp
```diff
@@ -0,0 +1,65 @@
...
+#ifndef VDA5050_CORE__ORDER_EXECUTION__ORDER_GRAPH_VALIDATOR_HPP_
+#define VDA5050_CORE__ORDER_EXECUTION__ORDER_GRAPH_VALIDATOR_HPP_
+
+#include <vector>
+
+#include "vda5050_types/edge.hpp"
+#include "vda5050_types/node.hpp"
+
+namespace vda5050_core {
+namespace order_graph_validator {
```
sauk2: Let's namespace it according to the folder structure, `vda5050_core::client::order` instead of `vda5050_core::order_graph_validator`

vda5050_core/src/vda5050_core/order_graph_validator/order_graph_validator.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+/// TODO (shawnkchan) maybe this should take in an Order object instead so that we can specify which orderId is errornous if something goes wrong
+/// TODO Refactor this to take in an Order object instead, will make validation logic easier
+OrderGraphValidator::OrderGraphValidator() {}
+
+bool OrderGraphValidator::is_valid_graph(
+  std::vector<vda5050_types::Node>& nodes,
+  std::vector<vda5050_types::Edge>& edges)
+{
+  /// TODO: (shawnkchan) Check if it is even possible to send an Order with zero nodes (should Order class just protect against this?). If not possible, get rid of this.
```
sauk2: It should not be possible to send an order with an empty node list. According to the spec, the order should have at least one node (the first entry in the list should always be the node the robot is currently on).

vda5050_core/include/vda5050_core/order_graph_validator/order_graph_validator.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+/// \brief Utility class with functions to perform validity checks on the graph contained in a VDA5050 Order message
```
sauk2: Could you try to keep the line numbers below 80 characters even for comments. It is the reason for the failing CI.

vda5050_core/src/vda5050_core/order_graph_validator/order_graph_validator.cpp
```diff
@@ -0,0 +1,152 @@
...
+  /// Does not seem to be any guarantee that this list is not empty, so will leave this check here.
+  if (nodes.empty())
+  {
+    std::cerr << "Graph Validation Error: nodes vector is empty.";
```
sauk2: I think it would be a better idea to use the VDA5050 logging macros

vda5050_core/include/vda5050_core/order_graph_validator/order_graph_validator.hpp
```diff
@@ -0,0 +1,65 @@
...
+{
+public:
+  OrderGraphValidator();
+
+  /// \brief Checks that the nodes and edges in a VDA5050 Order form a valid graph according to the VDA5050 specification sheet.
+  ///
+  /// \return True if nodes and edges create a valid graph, false otherwise
+  bool is_valid_graph(
+    std::vector<vda5050_types::Node>& nodes,
+    std::vector<vda5050_types::Edge>& edges);
```
sauk2: It would be a good idea to pass in the `vda5050_types::Order` class as a const reference instead of nodes and edges

vda5050_core/include/vda5050_core/order_graph_validator/order_graph_validator.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include "vda5050_types/node.hpp"
+
+namespace vda5050_core {
+namespace order_graph_validator {
+
+/// \brief Utility class with functions to perform validity checks on the graph contained in a VDA5050 Order message
+class OrderGraphValidator
+{
+public:
+  OrderGraphValidator();
```
sauk2: Right now the constructor does nothing so instead should we remove it and make all the functions static?

vda5050_core/include/vda5050_core/client/order/order_graph_validator.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+/// \brief Utility class with functions to perform validity checks on the graph contained in a VDA5050 Order message
+class OrderGraphValidator
+{
+public:
+  OrderGraphValidator();
+
+  /// \brief Checks that the nodes and edges in a VDA5050 Order form a valid graph according to the VDA5050 specification sheet.
+  ///
+  /// \return True if nodes and edges create a valid graph, false otherwise
```
sauk2, line 44: Returning a `bool` might not give sufficient information about the error itself as validation might fail due to many reasons. I would suggest making a new `ValidationResult` struct

```
struct ValidationResult
{
  bool valid;
  std::vector<vda5050_types::Error> errors;
}
```

The errors list is a mandatory field in the state message so the corresponding `errors` field can be directly assigned to the state message.

vda5050_core/include/vda5050_core/order_graph_validator/order_graph_validator.hpp
```diff
@@ -0,0 +1,65 @@
...
+  bool is_valid_graph(
+    std::vector<vda5050_types::Node>& nodes,
+    std::vector<vda5050_types::Edge>& edges);
+
+private:
+  // std::vector<vda5050_types::Node>& nodes;
+  // std::vector<vda5050_types::Edge>& edges;
+
+  std::string start_node_id;
+  std::string end_node_id;
```
sauk2: I feel like storing information like this might cause bugs later. Instead let's keep this class stateless as we have all the information we need in the `vda5050_types::Order` class.

sauk2 (changes requested):
Thanks for the great work! I have provided my first phase of review comments. 

I haven't read through the source file properly yet. These comments are mostly on the exposed API.

vda5050_core/CMakeLists.txt
shawnkchan: https://github.com/ros-industrial/vda5050_client/pull/13/commits/6179f8c9e14eac169809e0e272358021d805075b Moved `order_graph_validator.cpp` into `include/vda5050_core/client/order`

vda5050_core/include/vda5050_core/order_graph_validator/order_graph_validator.hpp
shawnkchan: https://github.com/ros-industrial/vda5050_client/pull/13/commits/6179f8c9e14eac169809e0e272358021d805075b Updated namespaces and header guards to follow `vda5050_core::client::order`

vda5050_core/src/vda5050_core/order_graph_validator/order_graph_validator.cpp
shawnkchan: https://github.com/ros-industrial/vda5050_client/pull/13/commits/0c8a8f04f65c98fc732a51871d98886ba116c084 Replaces `std::cerr` messages with VDA5050 logging macros

vda5050_core/include/vda5050_core/order_graph_validator/order_graph_validator.hpp
shawnkchan: https://github.com/ros-industrial/vda5050_client/pull/13/commits/0c8a8f04f65c98fc732a51871d98886ba116c084 Pass `vda50505_types::Order` as a const reference to `order_graph_manager`

vda5050_core/include/vda5050_core/order_graph_validator/order_graph_validator.hpp
shawnkchan: That makes a lot more sense
https://github.com/ros-industrial/vda5050_client/pull/13/commits/0c8a8f04f65c98fc732a51871d98886ba116c084 Remove constructor and make functions static

vda5050_core/include/vda5050_core/client/order/order_graph_validator.hpp
shawnkchan: https://github.com/ros-industrial/vda5050_client/pull/13/commits/a771210a315d3a31270f4a71666d0b4163812132 Create `ValidationResult` struct for more detailed error reporting

vda5050_core/include/vda5050_core/order_graph_validator/order_graph_validator.hpp
shawnkchan: https://github.com/ros-industrial/vda5050_client/pull/13/commits/0c8a8f04f65c98fc732a51871d98886ba116c084 make `OrderGraphValidator` stateless

vda5050_core/src/vda5050_core/order_graph_validator/order_graph_validator.cpp
shawnkchan: Removed the if statement checking for an empty node list

vda5050_core/src/vda5050_core/client/order/order_graph_validator.cpp
```diff
@@ -0,0 +1,194 @@
...
+#include <string>
+#include <variant>
+#include <vector>
+
+#include "vda5050_core/client/order/order_graph_validator.hpp"
+#include "vda5050_core/logger/logger.hpp"
+#include "vda5050_types/edge.hpp"
+#include "vda5050_types/error.hpp"
+#include "vda5050_types/error_level.hpp"
+#include "vda5050_types/node.hpp"
```
sauk2: Change to using angle brackets `<>` as these files are not present in the immediate search path.

vda5050_core/include/vda5050_core/client/order/order_graph_validator.hpp
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+
+#ifndef VDA5050_CORE__CLIENT__ORDER__ORDER_GRAPH_VALIDATOR_HPP_
+#define VDA5050_CORE__CLIENT__ORDER__ORDER_GRAPH_VALIDATOR_HPP_
+
+#include <vector>
+
+#include "vda5050_core/client/order/validation_result.hpp"
+#include "vda5050_types/order.hpp"
```
sauk2, line 25: Use angle brackets `<>` for this include 
```suggestion
#include <vda5050_types/order.hpp>
```

vda5050_core/test/unit/order_graph_validator/test_order_graph_validator.cpp
```diff
@@ -0,0 +1,124 @@
...
+ */
+
+#include <gtest/gtest.h>
+#include <optional>
+
+#include "vda5050_core/client/order/order_graph_validator.hpp"
+#include "vda5050_core/client/order/validation_result.hpp"
+#include "vda5050_types/edge.hpp"
+#include "vda5050_types/node.hpp"
+#include "vda5050_types/order.hpp"
```
sauk2: Change to angle brackets `<>`

vda5050_core/CMakeLists.txt
```diff
@@ -94,6 +113,19 @@ if(BUILD_TESTING)
     test/integration/mqtt_client/test_paho_mqtt_client.cpp
   )
   target_link_libraries(test_mqtt_client mqtt_client logger)
+
+  find_package(ament_cmake_gtest REQUIRED)
```
sauk2: I don't think you need to bring in `ament_cmake_gtest` separately. The previous `ament_cmake_gmock` most probably does that

vda5050_core/src/vda5050_core/client/order/order_graph_validator.cpp
```diff
@@ -0,0 +1,194 @@
...
+        "of the order's first and last node respectively.";
+      invalid_edges_error.error_level = vda5050_types::ErrorLevel::WARNING;
+
+      res.valid = false;
+      res.errors = {invalid_edges_error};
+      return res;
+    }
+  }
+  return res;
+}
```
sauk2: The current validation logic seems incorrect. Right now, the code compares every edge’s `start_node_id` only against the first node of the order and `end_node_id` only against the last node of the order. However, `start_node_id` and `end_node_id` on an edge describe the two nodes that the specific edge connects, not the global start and end of the entire order.

For a linear order, the correct relationship is sequence-based, i.e, the edge with sequence ID `n` should connect the node with sequence ID `n-1` (as its `start_node_id`) and the node with sequence ID `n+1` (as its `end_node_id`)

Because this validation depends entirely on the ordering of sequence IDs, it should not be done using only the first and last node. Instead, the check should be integrated into the part of the validation where you are already confirming that sequence IDs form a proper graph. That way you can directly verify that each edge connects the two adjacent nodes it is supposed to.

sauk2 (changes requested):
Some more comments regarding the functional logic. I feel there are a few validation checks still missing,

- All node sequence IDs should be even and all edge sequence IDs should be odd.
This follows from the structure of a valid order - an order must contain at least one node (the AGV’s current position), which always starts at sequence ID `0`. From there, nodes and edges should alternate.
- The validator should ensure that no two nodes or edges share the same sequence ID. Duplicate sequence IDs would break the ordering during execution.
- Sequence IDs should form a continuous chain with no gaps so there should be a check to search for any missing sequence IDs.
- There should be a clear boundary between base and horizon entities. Once the first unreleased node or edge is encountered, all following nodes and edges must also be unreleased.

johnaa13:
@sauk2 For review,

implements validation for VDA5050 order graphs to ensure basic order and sequencing rules are met before execution.

Features:
- is_valid_graph()
1) Enforces a linear chain where Nodes = Edges + 1.
2) Checks for parity (Nodes=Even, Edges=Odd) and continuity (no gaps in Sequence IDs)
3) Ensures "released" nodes never appear after "unreleased" nodes.
4) Ensure edge's start_node_id and end_node_id are correctly labelled.

- is_valid_order_update()
4) Logic to ensure new Order updates stitch correctly to the last released node of the base graph

vda5050_core/src/vda5050_core/client/order/order_graph_validator.cpp
```diff
@@ -16,339 +16,272 @@
...
+      std::vector<vda5050_types::ErrorReference> all_refs = {
+        {"order.order_id", order_id}, {"order.order_update_id", update_id}};
+      all_refs.insert(all_refs.end(), refs.begin(), refs.end());
+      VDA5050_ERROR("Order Validation Error: {}", msg);
+      res.errors.push_back(vda5050_types::Error{
+        "Order Validation Error", std::move(all_refs), msg,
+        vda5050_types::ErrorLevel::WARNING});
+    };
+
+  // check if there are exusting nodes, stop if empty
```
shawnkchan, line 49: typo on the word "exusting"

vda5050_core/src/vda5050_core/client/order/order_graph_validator.cpp
```diff
@@ -16,339 +16,272 @@
...
   }
-}
-
-ValidationResult OrderGraphValidator::is_valid_graph()
-{
-  ValidationResult res{true, {}};
 
-  /// number of nodes can only be 1 more than number of edges
-  if (!(order_.nodes.size() - 1 == order_.edges.size()))
+  // check if number of order is n, and number of edge is n-1
```
shawnkchan, line 56: check if number of nodes is n

vda5050_core/src/vda5050_core/client/order/order_graph_validator.cpp
```diff
@@ -16,339 +16,272 @@
...
-        current_edge.element_id);
-
-      vda5050_types::ErrorReference edge_error_reference{
-        "edgeId", current_edge.element_id};
-      invalid_edges_error.error_references = {edge_error_reference};
-
-      res.valid = false;
-      res.errors.push_back(invalid_edges_error);
-    }
+  // nnode must match (node_id and sequence_id)
```
shawnkchan, line 271: typo on "nnode"

vda5050_core/src/vda5050_core/client/order/order_graph_validator.cpp
```diff
@@ -16,339 +16,272 @@
...
+    add_error(
+      "The base (released graph) ends with an edge; it must end with a node.");
   }
+
   return res;
 }
 
-ValidationResult OrderGraphValidator::has_valid_sequence_ids()
+//=============================================================================
+ValidationResult OrderGraphValidator::is_valid_order_update(
```
shawnkchan, line 187: Hm I thought that the OrderManager would handle this. Could I clarify the rationale for having the OrderGraphValidator doing this check?

shawnkchan (commented):
Changes look good! Just spotted some typos in the comments that should be corrected.

sauk2:
Closing this PR in favour of #31

PR #14, opened 2025-11-25
Feat(state-manager): add StateManager

Johnarman1398:
Add StateManager implementation into vda5050_core

This PR implements tracking of VDA 5050-compliant state information.

### DIFF
--- a/vda5050_core/CMakeLists.txt
+++ b/vda5050_core/CMakeLists.txt
@@ -86,7 +86,12 @@ if(BUILD_TESTING)
   ament_cppcheck()
 
   find_package(ament_cmake_cpplint REQUIRED)
-  ament_cpplint()
+  set(cpplint_filters
+    "-whitespace/newline"
+  )
+  ament_cpplint(
+    FILTERS "${cpplint_filters}"
+  )
 
   find_package(ament_cmake_clang_format REQUIRED)
   ament_clang_format(
--- a/vda5050_core/CMakeLists.txt
+++ b/vda5050_core/CMakeLists.txt
@@ -34,10 +34,10 @@ target_include_directories(logger
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
 
-add_library(state_manager src/vda5050_core/state_manager/state_manager.cpp)
-target_link_libraries(state_manager
+add_library(client src/vda5050_core/state_manager/state_manager.cpp)
+target_link_libraries(client
   vda5050_types::vda5050_types)
-target_include_directories(state_manager
+target_include_directories(client
   PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
          $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 
@@ -46,7 +46,7 @@ target_link_libraries(vda5050_core
   INTERFACE
     mqtt_client
     logger
-    state_manager
+    client
 )
 target_include_directories(vda5050_core
   INTERFACE
@@ -63,7 +63,7 @@ install(
   TARGETS
     mqtt_client
     logger
-    state_manager
+    client
     vda5050_core
   EXPORT export_vda5050_core
   LIBRARY DESTINATION lib
@@ -109,7 +109,7 @@ if(BUILD_TESTING)
     test/unit/state_manager/test_state_manager.cpp
     test/integration/mqtt_client/test_paho_mqtt_client.cpp
   )
-  target_link_libraries(test_mqtt_client mqtt_client logger state_manager)
+  target_link_libraries(test_mqtt_client mqtt_client logger client)
 endif()
 
 ament_package()
--- a/vda5050_core/CMakeLists.txt
+++ b/vda5050_core/CMakeLists.txt
@@ -8,6 +8,7 @@ endif()
 find_package(ament_cmake REQUIRED)
 find_package(PahoMqttCpp REQUIRED)
 find_package(fmt REQUIRED)
+find_package(vda5050_types REQUIRED)
 
 include(GNUInstallDirs)
 
@@ -33,11 +34,19 @@ target_include_directories(logger
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
 
+add_library(state_manager src/vda5050_core/state_manager/state_manager.cpp)
+target_link_libraries(state_manager
+  vda5050_types::vda5050_types)
+target_include_directories(state_manager
+  PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+         $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
+
 add_library(vda5050_core INTERFACE)
 target_link_libraries(vda5050_core
   INTERFACE
     mqtt_client
     logger
+    state_manager
 )
 target_include_directories(vda5050_core
   INTERFACE
@@ -54,6 +63,7 @@ install(
   TARGETS
     mqtt_client
     logger
+    state_manager
     vda5050_core
   EXPORT export_vda5050_core
   LIBRARY DESTINATION lib
@@ -91,9 +101,10 @@ if(BUILD_TESTING)
     test/unit/mqtt_client/test_mqtt_client_interface.cpp
     test/unit/logger/test_logger.cpp
     test/unit/logger/test_default_logger.cpp
+    test/unit/state_manager/test_state_manager.cpp
     test/integration/mqtt_client/test_paho_mqtt_client.cpp
   )
-  target_link_libraries(test_mqtt_client mqtt_client logger)
+  target_link_libraries(test_mqtt_client mqtt_client logger state_manager)
 endif()
 
 ament_package()
--- a/vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
+++ b/vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
@@ -82,7 +82,7 @@ public:
 
   /// \brief get the current order_id
   /// \return order_id of the current order
-  std::optional<std::string> get_order_id() const;
+  std::string get_order_id() const;
 
   /// \brief set the current order_update_id
   /// \param order_update_id the current order update_id to set
@@ -90,7 +90,7 @@ public:
 
   /// \brief get the current order_update_id
   /// \return order_update_id of the current order
-  std::optional<uint32_t> get_order_update_id() const;
+  uint32_t get_order_update_id() const;
 
   /// \brief set the current zone_set_id
   /// \param zone_set_id zone_set_id the current order to set
@@ -166,7 +166,7 @@ public:
 
   /// \brief Get the current loads
   /// \return const std::vector<Load>& the current loads
-  const std::optional<std::vector<Load>>& get_loads() const;
+  std::optional<std::vector<Load>> get_loads() const;
 
   /// \brief Set the current operating mode of the AGV
   /// \param operating_mode the new operating mode
@@ -183,7 +183,7 @@ public:
 
   /// \brief Get the current battery state from the state
   /// \return const BatteryState& the current battery state
-  const BatteryState& get_battery_state() const;
+  BatteryState get_battery_state() const;
 
   /// \brief Set the current safety state of the AGV
   /// \param safety_state the safety state
@@ -192,7 +192,7 @@ public:
 
   /// \brief Get the current safety state from the state
   /// \return const SafetyState& the current safety state
-  const SafetyState& get_safety_state() const;
+  SafetyState get_safety_state() const;
 
   /// \brief Set the request new base flag
   /// \param request_new_base is AGV requesting for new base
@@ -250,7 +250,7 @@ public:
 
   /// @brief Get the current robot state.
   /// @return const State& the current robot state.
-  const State& get_state();
+  State get_state();
 };
 
 }  // namespace state_manager
--- a/vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
+++ b/vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
@@ -26,9 +26,11 @@
 #include <string>
 #include <vector>
 
+#include "vda5050_types/action_state.hpp"
 #include "vda5050_types/agv_position.hpp"
 #include "vda5050_types/battery_state.hpp"
 #include "vda5050_types/error.hpp"
+#include "vda5050_types/header.hpp"
 #include "vda5050_types/info.hpp"
 #include "vda5050_types/load.hpp"
 #include "vda5050_types/operating_mode.hpp"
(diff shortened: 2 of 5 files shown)


vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
```diff
@@ -0,0 +1,209 @@
...
+  /// \return The last reached node's ID.
+  std::string get_last_node_id() const;
+
+  /// \brief Get the sequenceId of the latest node the AGV has reached.
+  /// \return The last reached node's sequence ID.
+  uint32_t get_last_node_sequence_id() const;
+
+  /// \brief Set the current AGV position
+  /// \param agv_position the agv position
+  void set_agv_position(const std::optional<AGVPosition>& agv_position);
```
sauk2: Is there any specific reason you are passing in `agv_position` as an reference to optional? If not then you can just pass it as a `const` reference (`const AGVPosition& agv_position`). Since it is an optional field, the user will never call this setter unless they have some value to assign.

vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
```diff
@@ -0,0 +1,209 @@
...
+
+  /// \brief Get the current AGV position (if set)
+  /// \return std::optional<AGVPosition> the current AGV position of std::nullopt
+  std::optional<AGVPosition> get_agv_position() const;
+
+  /// \brief Set the current velocity
+  /// \param velocity the velocity
+  void set_velocity(const std::optional<Velocity>& velocity);
```
sauk2: Same comment as the `AGVPosition` setter

vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
```diff
@@ -0,0 +1,209 @@
...
+
+  /// \brief Get the Velocity (if set)
+  /// \return std::optional<Velocity> the velocity of std::nullopt
+  std::optional<Velocity> get_velocity() const;
+
+  /// \brief Set the driving flag of the AGV
+  /// \param driving is the agv driving?
+  /// \return true, if the driving flag changed
+  bool set_driving(bool driving);
```
sauk2: Since it is a basic setter, do we need to return a changed status here even if input value was the same?

vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
```diff
@@ -0,0 +1,209 @@
...
+  void reset_distance_since_last_node();
+
+  /// \brief Get the current distance since the last node.
+  /// \return The current distance since the last node, or std::nullopt if not set.
+  std::optional<double> get_distance_since_last_node() const;
+
+  /// \brief Add a new load to the state
+  /// \param load  the load to add
+  /// \return true if the loads changed (in this case always)
+  bool add_load(const Load& load);
```
sauk2: Should we change this to take in a vector of `Load` so that users can pass in multiple entries. I am not sure whether we should overwrite the entire vector or append to it. You can give me your thoughts on that

My concern is that it will be tedious to keep adding load values one by one

Also do we need to return a `bool`?

vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
```diff
@@ -0,0 +1,209 @@
...
+  bool remove_load(std::string_view load_id);
+
+  /// \brief Get the current loads
+  /// \return const std::vector<Load>& the current loads
+  const std::vector<Load>& get_loads();
+
+  /// \brief Set the current operating mode of the AGV
+  /// \param operating_mode the new operating mode
+  /// \return true if the operating mode changed
+  bool set_operating_mode(OperatingMode operating_mode);
```
sauk2: Same comment as `set_driving` function

vda5050_core/src/vda5050_core/state_manager/state_manager.cpp
```diff
@@ -0,0 +1,399 @@
...
+  return this->robot_state_.loads->size() != before_size;
+}
+
+//=============================================================================
+const std::vector<Load>& StateManager::get_loads()
+{
+  std::shared_lock lock(
+    this->mutex_);  // Ensure that loads is not being altered at the moment
+
+  static const std::vector<Load> empty;
```
sauk2: I am not sure if it's a good idea to manufacture a lifetime of an empty vector to avoid dangling references. Instead could we change the return type of this function to `const std::optional<std::vector<Load>>&` and mark the method as a `const`. We would be following the patterns that we have set in the previous functions.

vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
```diff
@@ -0,0 +1,209 @@
...
+  void set_battery_state(const BatteryState& battery_state);
+
+  /// \brief Get the current battery state from the state
+  /// \return const BatteryState& the current battery state
+  const BatteryState& get_battery_state() const;
+
+  /// \brief Set the current safety state of the AGV
+  /// \param safety_state the safety state
+  /// \return true if the state changed
+  bool set_safety_state(const SafetyState& safety_state);
```
sauk2: Same comment as `set_driving` function

vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
```diff
@@ -0,0 +1,209 @@
...
+
+  /// \brief Get the current safety state from the state
+  /// \return const SafetyState& the current safety state
+  const SafetyState& get_safety_state() const;
+
+  /// \brief Set the request new base flag to true
+  void request_new_base();
```
sauk2: This sets the `newBaseRequest` value to `true` but we need to find a way to make it `false` upon receiving a new order otherwise it will be sent with each state update.

vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
```diff
@@ -0,0 +1,209 @@
...
+
+  /// \brief Add an error to the state
+  /// \param error the error to add
+  /// \return true if the errors array changed (in this case always)
+  bool add_error(const Error& error);
```
sauk2: Same comments as `add_load(..)`

vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
```diff
@@ -0,0 +1,209 @@
...
+
+  /// \brief Get a copy of the current errors
+  /// \return std::vector<Error>
+  std::vector<Error> get_errors() const;
+
+  /// \brief Add a new information to the state
+  /// \param info the information to add
+  void add_info(const Info& info);
```
sauk2, line 217: Same comments as `add_load(..)`

vda5050_core/CMakeLists.txt
```diff
@@ -33,11 +34,19 @@ target_include_directories(logger
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
 
+add_library(state_manager src/vda5050_core/state_manager/state_manager.cpp)
```
sauk2: Let's change the library name to `client` that includes all the state and order related features. The order related code will be added to the same library

vda5050_core/src/vda5050_core/state_manager/state_manager.cpp
```diff
@@ -0,0 +1,399 @@
...
+  nodes.erase(
+    std::remove_if(nodes.begin(), nodes.end(), node_predicate), nodes.end());
+  edges.erase(
+    std::remove_if(edges.begin(), edges.end(), edge_predicate), edges.end());
+}
+
+void StateManager::append_states_for_update(Order& order_update)
+{
+  this->set_new_order(order_update);
+}
```
sauk2: Do we need this function when all it is doing is forwarding calls to `set_new_order`? The "append" keyword gives an impression that something is being modified but actually the whole `robot_state_` variable is being overwritten.

sauk2 (changes requested):
Thanks for PR! I have provided some initial review comments.

I feel there are some features still missing from the `StateManager` class,
- Getters and setters for the `action_states` variable
- Function to set the header. I would expect the users to directly serialize the `vda5050_types::State` class into `nlohmann::json` and in the current implementation `get_state` would be missing the header fields.
- Function to update the order progress. There is no way to update the `last_node_id` and `last_node_sequence_id` fields. In our current design, these fields will be read by the `OrderManager` to increment the order and dispatch the next node/edge.

I also have a general comment on the API design. Most of the get functions return references to fields inside the internal `robot_state_` object. This works only under the assumption that the `StateManager` outlives all consumers of those references and callers never store those references beyond the `StateManager`’s lifetime.

If the `StateManager` were ever destroyed earlier than expected, any stored references would become dangling and could lead to undefined behavior as it tightly couples the lifetime of external code to the lifetime of `StateManager`.

We should discuss whether returning values would make the API safer, especially if the `StateManager` is ever used without our execution engine. I would also like to include @Briancbn in this discussion for his thoughts!

sauk2:
@Johnarman1398 you can ignore the CI failures for now. The failing tests are due to failure to connect to the online [MQTT broker](https://test.mosquitto.org/). I will try to investigate this issue.

sauk2:
> I will try to investigate this issue.

The issue seems to have solved itself XD

vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
```diff
@@ -0,0 +1,259 @@
...
+  /// \brief Set a new order on the vehicle (after clearing any existing order).
+  /// \param order The new order to accept and store.
+  void set_new_order(const Order& order);
+
+  /// \brief Clear the horizon nodes/edges from the current nodeStates and edgeStates.
+  void clear_horizon();
+
+  /// @brief Get the current robot state.
+  /// @return const State& the current robot state.
+  const State& get_state();
```
sauk2: Let's return this also by value instead of const reference

vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
```diff
@@ -0,0 +1,259 @@
...
+  /// \return OperatingMode the current operating mode
+  OperatingMode get_operating_mode() const;
+
+  /// \brief Set the current battery state of the AGV
+  /// \param battery_state the battery state
+  void set_battery_state(const BatteryState& battery_state);
+
+  /// \brief Get the current battery state from the state
+  /// \return const BatteryState& the current battery state
+  const BatteryState& get_battery_state() const;
```
sauk2: Change to return by value

vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
```diff
@@ -0,0 +1,259 @@
...
+
+  /// \brief Set the current safety state of the AGV
+  /// \param safety_state the safety state
+  /// \return true if the state changed
+  void set_safety_state(const SafetyState& safety_state);
+
+  /// \brief Get the current safety state from the state
+  /// \return const SafetyState& the current safety state
+  const SafetyState& get_safety_state() const;
```
sauk2: Change to return by value

vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
```diff
@@ -0,0 +1,259 @@
...
+  void set_load(const std::vector<Load>& loads);
+
+  /// \brief Remove a load by it's id from the loads array
+  /// \param load_id the id of the load to remove
+  /// \return true if at least one load was removed
+  bool remove_load(std::string_view load_id);
+
+  /// \brief Get the current loads
+  /// \return const std::vector<Load>& the current loads
+  const std::optional<std::vector<Load>>& get_loads() const;
```
sauk2: Change to return by value

vda5050_core/src/vda5050_core/state_manager/state_manager.cpp
```diff
@@ -0,0 +1,487 @@
...
+{
+  std::unique_lock lock(this->mutex_);
+  this->robot_state_.zone_set_id = zone_set_id;
+}
+
+//=============================================================================
+std::optional<std::string> StateManager::get_zone_set_id() const
+{
+  std::shared_lock lock(this->mutex_);
+  if (this->robot_state_.order_id.empty()) return std::nullopt;
```
sauk2: From my understanding of the spec, `zone_set_id` seems to be independent of the order itself. 

Image

vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
```diff
@@ -0,0 +1,259 @@
...
+  /// \return header of the AGV
+  Header get_header() const;
+
+  /// \brief set the current order_id
+  /// \param order_id of the current order
+  void set_order_id(const std::string& order_id);
+
+  /// \brief get the current order_id
+  /// \return order_id of the current order
+  std::optional<std::string> get_order_id() const;
```
sauk2: Since `order_id` is a required field, I am not sure if it's a good idea to return an `std::optional`. I think returning the field itself (empty or not) should be fine and then the user of this class can decide what to do with that information.

vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
```diff
@@ -0,0 +1,259 @@
...
+
+  /// \brief set the current order_update_id
+  /// \param order_update_id the current order update_id to set
+  void set_order_update_id(uint32_t order_update_id);
+
+  /// \brief get the current order_update_id
+  /// \return order_update_id of the current order
+  std::optional<uint32_t> get_order_update_id() const;
```
sauk2: Same comment as that for return type of `order_id` goes for this too.

vda5050_core/include/vda5050_core/state_manager/state_manager.hpp
sauk2, line 1: We can follow this folder structure - `include/vda5050_core/client/state/state_manager.hpp` and `src/vda5050_core/client/state/state_manager.cpp`  and corresponding namespacing `vda5050_core::client::state::`.

Similar approach is being followed in the order-related PRs

vda5050_core/src/vda5050_core/state_manager/state_manager.cpp
```diff
@@ -0,0 +1,487 @@
...
+  for (const auto& action_state : this->robot_state_.action_states)
+  {
+    if (
+      action_state.action_status != ActionStatus::FINISHED &&
+      action_state.action_status != ActionStatus::FAILED)
+    {
+      return true;
+    }
+  }
+  return true;
```
sauk2, line 392: I am guessing this must return `false` by default

sauk2 (changes requested):
Thanks for the changes. I think this PR is almost ready to be merged!

vda5050_core/src/vda5050_core/state_manager/state_manager.cpp
johnaa13: thanks for spotting this, completely missed this haha

vda5050_core/CMakeLists.txt
sauk2: I think you missed this change

sauk2 (approved):
Looks good! I will merge it in once you make that remaining change

PR #19, opened 2025-12-10
feat(mqtt): update mqtt interface with new methods

sauk2:
## Summary

This PR adds 2 new methods to the MQTT interface and modifies an existing method. It also updates the Paho client and unit tests to reflect the changes.

New methods added,
- `connected()` - Check MQTT connection status
- `set_will(...)` - Set a last will message that will be sent when the client disconnects abruptly

Existing methods updated,
- `publish(...)` - A new `boolean` flag has been added to allow users to indicate if this message should be retained in the broker

### DIFF
--- a/vda5050_core/include/vda5050_core/mqtt_client/mqtt_client_interface.hpp
+++ b/vda5050_core/include/vda5050_core/mqtt_client/mqtt_client_interface.hpp
@@ -48,7 +48,8 @@ public:
   /// \param message Raw message string
   /// \param qos Quality of service setting for the publish
   virtual void publish(
-    const std::string& topic, const std::string& message, int qos) = 0;
+    const std::string& topic, const std::string& message, int qos,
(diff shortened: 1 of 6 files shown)


tanjpg (approved):
LGTM. Not sure if you want to include tests for testing that the last will is indeed broadcasted properly when client disconnects after you set the specific `set_will` method.

sauk2:
> LGTM. Not sure if you want to include tests for testing that the last will is indeed broadcasted properly when client disconnects after you set the specific `set_will` method.

Thanks for catching this! Verifying that the will is actually broadcast on abrupt disconnect is difficult to test reliably in code because destroying the underlying Paho instance invokes its destructor which invokes a safe disconnect. As an alternative, I exposed the internal Paho configuration and validate that the will is configured correctly by checking that the client initializes successfully and that the will topic and payload are set as expected.

PR #20, opened 2025-12-10
Feat (VDA5050 master) : Initial commit of VDA5050 Master Codebase

tanjpg:
This PR contains the commits for the current initial status of the VDA5050 master prototype package. Future PRs to this package will be incremental. 

# Features

## 1.  Abstracted communication layer

The [ICommunicationStrategy](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/include/vda5050_master/communication/communication.hpp#L55) class contains a generic class for communication from the VDA5050 Master to the VDA5050 AGV.
We use the generic [ConnectionState](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/include/vda5050_master/communication/communication.hpp#L47) enum to represent the connection states.

There are currently 2 implementations of this class. 

### 1a. MqttCommunication (Inherits`ICommunicationStrategy`)
[MqttCommunication](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/include/vda5050_master/communication/mqtt.hpp#L29) contains the MQTT Implementation of the `ICommunicationStrategy` class. This uses the [`mqtt_client`](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_core/include/vda5050_core/mqtt_client/mqtt_client_interface.hpp) implementation of the mqtt client as the base API layer

Unit tests for this class [here](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/test/communication/mqtt/communications.cpp)

### 1b. Ros2Communication (Inherits`ICommunicationStrategy`)
[Ros2Communication](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/include/vda5050_master/communication/ros2.hpp#L31) contains standard implementation of ROS2 related pub-sub architecture.

Unit tests for this class [here](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/test/communication/ros2/communications.cpp)

## 2. Heartbeat implementation
As per the VDA5050 specifications, the `connection` and `state` topics require a certain minimum frequency of publishing. The [HeartbeatListener](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/include/vda5050_master/communication/heartbeat.hpp#L49) class is the generic listener that will take callbacks to handle the state transition for the AGV given `connection` or `state` message lapses.

Unit tests for this class [here](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/test/integration/heartbeat.cpp)

## 3. AGV class implementation
This current design uses the [AGV](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/include/vda5050_master/agv/agv.hpp#L77) class with its own individual `ICommunicationStrategy` pubs and subs. these subs will handle two different states. The [AGVConnectionState](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/include/vda5050_master/agv/agv.hpp#L47) enum handles the `connection` message related state transitions, while the [AGVState](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/include/vda5050_master/agv/agv.hpp#L57) enum handles `state` message related state transitions. These topic subs will have separate `HeartbeatListener`s

Unit tests for class [here](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/test/agv/agv_test.cpp)

### 3a. AGVConnectionState state transition

When AGV is first onboarded, or when disconnected explicitly- `OFFLINE`

When no `connection` message received after seconds - 'CONNECTIONBROKEN`

When receiving online status - `ONLINE`

Unit tests for state transitions [here](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/test/agv/agv_connection_state_test.cpp)

### 3.b AGVState state transition

When AGV is first onboarded - `STATE_UNKNOWN`

When `state` message comes in - `AVAILABLE`

When `state` message does not come in and heartbeat fails - `STATE_UNKNOWN`

If ever `AGVConnectionState` transitions to `OFFLINE` or `CONNECTIONBROKEN` - `UNAVAILABLE`

Any errors reported in `state` message - `ERROR`

Unit tests for this state transition [here](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/test/agv/agv_operational_state_test.cpp)

## 4. VDA5050 Master
An initial skeleton of how (1) to (3) ties in together is shown in the [VDA5050Master](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/include/vda5050_master/vda5050_master/master.hpp) class. The main thing added here is the onboarding and offboarding methods ([`register_agv`](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/include/vda5050_master/vda5050_master/master.hpp#L107) and [`unregister_agv`](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/include/vda5050_master/vda5050_master/master.hpp#L119)). Simple functions for publishing to `order` and `instant actions` are also added.

Unit tests for this class is also in the same file as the one for `AGV` [here](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/test/agv/agv_test.cpp) Since the main functions to test are mainly `registering` and `unregistering` AGVs

In future, the main VDA5050 workflow logic will be added to this class.


vda5050_master/include/vda5050_master/agv/agv.hpp
```diff
@@ -0,0 +1,376 @@
...
+ *
+ * Maps directly to the connectionState field in VDA5050 connection messages.
+ * Values: ONLINE, OFFLINE, CONNECTIONBROKEN
+ */
+enum class AGVConnectionState
+{
+  ONLINE,   // AGV reports connectionState: "ONLINE"
+  OFFLINE,  // AGV reports connectionState: "OFFLINE"
+  CONNECTIONBROKEN  // AGV reports connectionState: "CONNECTIONBROKEN" or heartbeat timeout
+};
```
sauk2: I think the `vda5050_types::ConnectionState` enum class can be reused for this

vda5050_master/include/vda5050_master/vda5050_interfaces.hpp
sauk2, line 1: Could we replace `vda5050_msgs` with `vda5050_types`? I think those classes would fit better here

vda5050_master/include/vda5050_master/standard_names.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+const int ConnectionQos = 1;
+const int FactsheetQos = 0;
+const int OrderQos = 1;
+const int StateQos = 1;
+const int VisualizationQos = 0;
+const int InstantActionsQos = 1;
+
+const int ConnectionHeartbeatInterval = 15;  // seconds
+const int StateHeartbeatInterval = 30;       // seconds
```
sauk2, line 42: The [`factsheet.schema`](https://github.com/VDA5050/VDA5050/blob/5250525e602f5ae9ad4cc5da6948112de981a4c0/json_schemas/factsheet.schema#L291-L322) has an entry for this. Is it instead a better idea to request for factsheet from the AGV once a connection message is received?

vda5050_master/include/vda5050_master/agv/agv.hpp
```diff
@@ -0,0 +1,376 @@
...
+  void handle_visualization_message(
+    const std::string& payload, const VisualizationCallback& callback);
+
+  // Identity
+  std::string manufacturer_;
+  std::string serial_number_;
+  std::string agv_id_;
+
+  // Communication
+  std::unique_ptr<ICommunicationStrategy> communication_;
```
sauk2: I see that every AGV will have a new client instance. Is it instead possible to house the communication outside the AGV instance and instead of creating multiple MQTT client instances, we have multiple subscriptions? Or what about having one subscription using wildcards (eg. `+/v2/+/+/connection`) and then sending to appropriate AGV instance by either reading individual subtopics or the header?

vda5050_master/include/vda5050_master/agv/agv.hpp
```diff
@@ -0,0 +1,376 @@
...
+  mutable std::mutex state_mutex_;
+  AGVConnectionState connection_status_{AGVConnectionState::OFFLINE};
+  AGVState operational_state_{AGVState::STATE_UNKNOWN};
+
+  // Internal state setters (called by heartbeat callbacks)
+  void set_connection_status(AGVConnectionState status);
+  void set_operational_state(AGVState state);
+
+  // Heartbeat timeout callbacks
+  void on_connection_heartbeat_timeout();
```
sauk2: Image

Although a 15 second heartbeat is mentioned here, I think this can be handled as a broker configuration. The way this mechanism has been implemented in the AGV side is by sending a first `ONLINE` message with QoS 1 and a retained flag to `true`. That way the message is stored in the broker for the foreseeable future so even if you poll the broker for the connection message, you will keep getting an old message. A last will is set in the AGV side so that if the broker cannot poll the client, it will automatically send a `CONNECTIONBROKEN` message to the upstream listener. At the end during disconnection, an `OFFLINE` message is sent before calling `disconnect()` on the client. So, I was thinking on the master side you can take advantage of this by receiving a connection message at the start which will stay in the broker forever and any successive messages either due to broken connection (sent by the broker itself) or graceful disconnection (sent from the client) can be tracked in the subscription callback.


Image

Based on the successive paragraphs, I see that all messages are supposed to be set to retained so I would assume there is no requirement for manual polling with a heartbeat because it will be the same message always. Any new message will show up in the subscription callback.

vda5050_core/src/vda5050_core/mqtt_client/paho_mqtt_client.cpp
```diff
@@ -123,6 +123,7 @@ void PahoMqttClient::connect()
   {
     VDA5050_ERROR_STREAM(
       "Unable to establish MQTT connection: " << e.get_message());
+    throw;
```
sauk2: There is a PR (https://github.com/ros-industrial/vda5050_client/pull/19) that adds a new `connected()` API to check the connection status. The `connect()` function is blocking, so calling a `connected()` immediately after `connect()` returns will let us know if there was any failure. I feel that would be a better approach than throwing an error

vda5050_master/include/vda5050_master/standard_names.hpp
```diff
@@ -0,0 +1,45 @@
...
+const std::string StateTopic = "state";                     // NOLINT
+const std::string InstantActionsTopic = "instant_actions";  // NOLINT
+const std::string VisualizationTopic = "visualization";     // NOLINT
+
+const int ConnectionQos = 1;
+const int FactsheetQos = 0;
+const int OrderQos = 1;
+const int StateQos = 1;
+const int VisualizationQos = 0;
+const int InstantActionsQos = 1;
```
sauk2: Image

From reading 2.0.0 and 2.1.0 version of the spec, all topics except connection should have QoS 0. Although in 2.0.0 I cannot find any QoS mention for `instantActions`, I feel it is safe to assume it should be 0 based on 2.1.0 (attached image)

vda5050_master/include/vda5050_master/communication/communication.hpp
sauk2, line 1: I feel as a first phase we could try to fix on a traditional master control architecture (one that talks on MQTT), as described in the VDA5050 specification document. Later we could deliberate over how to achieve communication over ROS, whether we can do it just as an external bridge or if it is more useful to tightly couple with this repo. 

So, I was thinking just keeping the MQTT communication for now and instead of having an interface that has overlap with the `MqttClientInterface`, we could store a `shared_ptr` to handle communication.

sauk2 (changes requested):
Thanks for the PR! It is definitely useful to ship this package with a master control that can be used out of the box with minimal configurations!

I think it is time to evaluate a good name for this repo since it will more than just client related functionality now 

I have added some of comments from my understanding of the VDA5050 document

vda5050_master/include/vda5050_master/vda5050_interfaces.hpp
tanjpg: Addressed in [This Commit](https://github.com/ros-industrial/vda5050_client/pull/20/commits/f60d580e41d85c5eafe0742be08cf14d75c9bb46)

vda5050_master/include/vda5050_master/agv/agv.hpp
tanjpg: Addressed in [this commit](https://github.com/ros-industrial/vda5050_client/pull/20/commits/f60d580e41d85c5eafe0742be08cf14d75c9bb46). Removed existing enum

tanjpg:
Additionally, with the new addition of the [dummy methods](https://github.com/ros-industrial/vda5050_client/pull/24), I have [commented out unit tests that make use of the output from such methods](https://github.com/ros-industrial/vda5050_client/pull/20/commits/12b91a38a0990369d42be4b4483025d74cf88ee5)

vda5050_master/include/vda5050_master/standard_names.hpp
tanjpg: I think this can only be done properly once the factsheet parser is done. Suggest to move it to another PR?

vda5050_core/src/vda5050_core/mqtt_client/paho_mqtt_client.cpp
tanjpg: Fixed in [this commit](https://github.com/tanjpg/vda5050_client/commit/3e0526e0a66afbb0be0a9560c366230d9ee5167a)

vda5050_master/include/vda5050_master/standard_names.hpp
tanjpg: Fixed in [this commit](https://github.com/tanjpg/vda5050_client/commit/3e0526e0a66afbb0be0a9560c366230d9ee5167a)

vda5050_master/include/vda5050_master/agv/agv.hpp
tanjpg: Have rectified this behavior in a few ways:

1. [Removed the `connection` topic heartbeat](https://github.com/ros-industrial/vda5050_client/commit/9047b98fc25600283bacd98f99eb6d4ce6fa0435)
2. The `connection` topic will now listen for an `ONLINE` connection status, after which it calls the [`setup_agv_components`](https://github.com/ros-industrial/vda5050_client/commit/aac42c1a7a1e4efc125ad098b98fbbcdc492f938) method to construct the other sub/pub
3. Upon receiving the `OFFLINE` or `CONNECTIONBROKEN` status, it calls the [`cleanup_agv_components`](https://github.com/ros-industrial/vda5050_client/commit/c87c0f9608bca887df1c721537c37261dfbb4b0f) method. 

For now, since we are not using any database, the last messages are kept as-is, allowing for restarting the AGV components via a new `ONLINE` message.

vda5050_master/include/vda5050_master/communication/communication.hpp
Briancbn: I think the communication interface here serves the same purpose as the `MqttClientInterface` in `vda5050_core`. If we want to add ROS2 support the same way here, I would suggest to 
- rename `MqttClientInterface` into `ClientInterface`
- Introduce observer to the `ClientInterface` to allow handling of `on_connect` etc.. We can then provide the observer here to get the connection state or print out trace log if needed.
- Then we implement `ROS2Client`, if needed.

tanjpg: I think for this phase, as @sauk2 mentioned, I will be ok to keep the interface to be the `MqttClientInterface` first. There are potential wider concerns with using ROS2 and other communication methods (like wildcarding for example) that may require more thought when designing this abstract interface. We can just follow the official documentation's way as-is for now

tanjpg: Have reverted the use of `ICommunicationStrategy` to solely use `MqttClientInterface` [here](https://github.com/tanjpg/vda5050_client/commit/f2d777f98451effe893a2810aa073f8e53bd815b)

vda5050_master/include/vda5050_master/agv/agv.hpp
tanjpg: Have made the major changes in [this commit](https://github.com/tanjpg/vda5050_client/commit/55770b2610147d7dfbca3ac1202a64cc155f8314)

In summary:

1. [One MQTT Client](https://github.com/tanjpg/vda5050_client/blob/55770b2610147d7dfbca3ac1202a64cc155f8314/vda5050_master/include/vda5050_master/vda5050_master/master.hpp#L309) in the master subscribing to wildcarded topics

2. [Templated subscription callbacks](https://github.com/tanjpg/vda5050_client/blob/55770b2610147d7dfbca3ac1202a64cc155f8314/vda5050_master/src/vda5050_master/master.cpp#L358) to funnel information to the relevant `AGV` object

3. [Onboarding](https://github.com/tanjpg/vda5050_client/blob/55770b2610147d7dfbca3ac1202a64cc155f8314/vda5050_master/src/vda5050_master/master.cpp#L209) and [Offboarding](https://github.com/tanjpg/vda5050_client/blob/55770b2610147d7dfbca3ac1202a64cc155f8314/vda5050_master/src/vda5050_master/master.cpp#L232) methods to govern which topic messages will be dropped.

4. On the AGV side, transient mqtt clients will be created to [publish orders](https://github.com/tanjpg/vda5050_client/blob/55770b2610147d7dfbca3ac1202a64cc155f8314/vda5050_master/src/agv/agv.cpp#L483)  and [instant actions](https://github.com/tanjpg/vda5050_client/blob/55770b2610147d7dfbca3ac1202a64cc155f8314/vda5050_master/src/agv/agv.cpp#L521) that will be disconnected and destroyed at the end of the method.

vda5050_master/include/vda5050_master/agv/agv.hpp
```diff
@@ -0,0 +1,347 @@
...
+ * handles outgoing messages via transient MQTT clients.
+ *
+ * Thread safety: Methods are thread-safe. Cached data access is protected
+ * by mutexes.
+ */
+class AGV
+{
+public:
+  // Type aliases
+  using Clock = std::chrono::steady_clock;
```
sauk2: I think this can be changed to `std::chrono::system_clock` to keep in line with the chrono type used in `header.hpp`. From my understanding, `steady_clock` is to be used as a stopwatch as it only progresses forward while `system_clock` is to be used for date and time related stuff.

vda5050_master/include/vda5050_master/agv/agv.hpp
```diff
@@ -0,0 +1,347 @@
...
+   * @param serial_number Serial number
+   * @param broker_address MQTT broker address for creating transient publish clients
+   * @param max_queue_size Maximum number of outgoing messages to queue (default: 10)
+   * @param drop_oldest If true, drop oldest message when queue full; if false, reject new (default: true)
+   * @param state_heartbeat_interval State heartbeat timeout in seconds
+   */
+  AGV(
+    const std::string& manufacturer, const std::string& serial_number,
+    const std::string& broker_address,
+    size_t max_queue_size = DEFAULT_MAX_QUEUE_SIZE, bool drop_oldest = true,
```
sauk2: Instead of creating a static variable, would it be a better idea to add a default value in the constructor?


```suggestion
    size_t max_queue_size = 10, bool drop_oldest = true,
```

vda5050_master/src/agv/agv.cpp
```diff
@@ -0,0 +1,552 @@
...
+  {
+    // Create transient client
+    std::string client_id =
+      agv_id_ + "_actions_" +
+      std::to_string(Clock::now().time_since_epoch().count());
+    auto client = vda5050_core::mqtt_client::create_default_client(
+      broker_address_, client_id);
+
+    // Connect, publish, disconnect
+    client->connect();
```
sauk2: I think creating a new client with a unique ID might induce some latency in the system as each new MQTT client creation would require some form of handshake and authentication. 

Another concern I see is in terms of scalability since the performance won't be dependent just on the number of active AGVs in the system anymore but on how the upstream scheduling system behaves. For example. if the master handles 100 AGVs and each AGV is being sent 3 instant actions at an arbitrary timestep, that would create 300 new MQTT clients to the broker within a short time!

So instead, do you think it is a good idea to create one MQTT client per AGV, at the start just for the purpose of publishing? 

After we do that, we could also investigate if it is better to keep the publish clients alive or make them inactive by calling a `disconnect()` after every publish and calling a `connect()` before making a publish. This would remove the strain of handshake and authentication as the broker would probably have that client information stored in memory after calling of the first `connect()`.

vda5050_master/src/agv/agv.cpp
```diff
@@ -0,0 +1,552 @@
...
+
+    nlohmann::json j;
+    vda5050_types::to_json(j, order);
+    client->publish(
+      build_topic(vda5050_master::OrderTopic), j.dump(),
+      vda5050_master::OrderQos);
+
+    client->disconnect();
```
sauk2: Same concern as the one highlighted in `publish_instant_actions(...)`

vda5050_master/src/agv/agv.cpp
```diff
@@ -0,0 +1,552 @@
...
+  VDA5050_INFO("[AGV] Destroying AGV instance: {}", agv_id_);
+
+  // Stop queue processor
+  stop_queue_processor();
+
+  // Cleanup heartbeat
+  cleanup_heartbeat();
+
+  VDA5050_INFO("[AGV] AGV instance destroyed: {}", agv_id_);
+}
```
sauk2, line 67: What do you think about providing a `stop()` method that does this instead of closing everything inside the destructor?

If a user decides to use the class as is, they will need to destroy the class even if they want to pause execution for some reason and recreate it again if they want to bring it online.

sauk2 (commented):
Thanks for the changes! I have provided some inline comments, one specifically regarding the transient publish clients

vda5050_master/include/vda5050_master/agv/agv.hpp
tanjpg: Addressed [here](https://github.com/ros-industrial/vda5050_client/pull/20/commits/baec55a6525c4c06e048701d2c11b87ad178cf2b)

vda5050_master/include/vda5050_master/agv/agv.hpp
tanjpg: Addressed [here](https://github.com/ros-industrial/vda5050_client/pull/20/commits/baec55a6525c4c06e048701d2c11b87ad178cf2b)

vda5050_master/src/agv/agv.cpp
tanjpg: Addressed [here](https://github.com/ros-industrial/vda5050_client/pull/20/commits/f78cf223c178e113745c2277e58ce45168eb2e8b)

Along with `stop`, I have also additionally added `restart`, `pause` and `resume`

- `stop` mainly stops all the queue processing and heartbeat checking, clears the queue , and sets all the statuses to offline
- `restart` does similar things as `stop` but also clears the values of the AGV object.
- `pause` stops the queue processing and heartbeat checking but does not clear the queue.
- `resume` continues the queue processing and starts up the heartbeat checking

vda5050_master/src/agv/agv.cpp
tanjpg: Addressed [here](https://github.com/ros-industrial/vda5050_client/pull/20/commits/0da080a8cf4b0b3bef9c3ee582338273fab59fad). As discussed we will use a single mqtt client per agv object, that will use `connect` and `disconnect` to publish order and instant actions. the client id will be the unique id of the agv.

vda5050_master/src/agv/agv.cpp
```diff
@@ -0,0 +1,597 @@
...
+    return;  // Already set up
+  }
+
+  VDA5050_INFO("[AGV] Setting up heartbeat for {}", agv_id_);
+
+  state_heartbeat_ =
+    std::make_unique<vda5050_master::communication::HeartbeatListener>(
+      agv_id_ + "_state_heartbeat", state_heartbeat_interval_,
+      [this]() { on_state_heartbeat_timeout(); });
+  state_heartbeat_->start_connection_heartbeat();
```
sauk2: I think there is a potential for a race condition here with `state_heartbeat_`. Maybe better to protect it with a mutex?

vda5050_master/src/agv/agv.cpp
```diff
@@ -0,0 +1,597 @@
...
+  {
+    return;  // Not running
+  }
+
+  VDA5050_INFO("[AGV] Stopping queue processor for {}", agv_id_);
+
+  {
+    std::lock_guard<std::mutex> lock(queue_mutex_);
+    stop_processing_ = true;
+  }
```
sauk2: `stop_processing_` is atomic so probably not necessary to guard it with a mutex.

vda5050_master/include/vda5050_master/agv/agv.hpp
```diff
@@ -0,0 +1,383 @@
...
+  bool drop_oldest_;
+
+  std::mutex queue_mutex_;
+  std::condition_variable queue_cv_;
+  std::queue<vda5050_types::Order> order_queue_;
+  std::queue<vda5050_types::InstantActions> instant_actions_queue_;
+
+  // Queue processing thread
+  std::atomic<bool> stop_processing_{false};
+  std::atomic<bool> queue_processor_running_{false};
```
sauk2: Can we use the type alias `std::atomic_bool` instead as we are using it in other packages?

vda5050_master/src/agv/agv.cpp
```diff
@@ -0,0 +1,597 @@
...
+               !instant_actions_queue_.empty();
+      });
+
+      // Check if we should stop
+      if (
+        stop_processing_ && order_queue_.empty() &&
+        instant_actions_queue_.empty())
+      {
+        break;
+      }
```
sauk2, line 546: This would mean that the queues will still be emptied even after `stop_processing_` becomes `true`. Should orders/instant actions keep getting sent till their respective queues are empty or should they be dropped?

vda5050_master/src/agv/agv.cpp
```diff
@@ -0,0 +1,597 @@
...
+  if (queue_processor_running_.load())
+  {
+    return;  // Already running
+  }
+
+  VDA5050_INFO("[AGV] Starting queue processor for {}", agv_id_);
+
+  stop_processing_ = false;
+  queue_processor_running_ = true;
+  queue_thread_ = std::thread(&AGV::process_queues, this);
```
sauk2, line 487: I think there is a risk of double-starting the queue thread here. Although `queue_processor_running_` is atomic, the function as a whole is not thread-safe.

For example, if two threads call this function concurrently, both may observe `queue_processor_running_ == false`. The first thread then sets it to `true` and starts the queue thread, followed by the second thread doing the same.

While the atomic guarantees that the variable itself is read and written safely, it does not make the entire sequence atomic, so the subsequent operations are not protected. So maybe adding a queue mutex can resolve this?

vda5050_master/include/vda5050_master/communication/heartbeat.hpp
```diff
@@ -0,0 +1,249 @@
...
+
+  void start_connection_heartbeat()
+  {
+    {
+      std::lock_guard<std::mutex> lock(state_mutex_);
+      VDA5050_INFO("Starting Connection heartbeat listener");
+      state_ = HeartbeatState::RUNNING;
+    }
+    connection_thread_ = std::thread(&HeartbeatListener::listen, this);
+  }
```
sauk2: This function is indirectly bound to the connection subscription.  There is a risk that if an AGV sends multiple `ONLINE` messages, then each time a new thread will start and the already started thread will not be joinable. A fix could be to check if this heartbeat is already running before creating a new thread.

vda5050_master/include/vda5050_master/communication/heartbeat.hpp
```diff
@@ -0,0 +1,249 @@
...
+        // Copy callback by value to ensure safe invocation
+        auto callback_copy = disconnection_callback_;
+        callback_thread_ = std::thread([callback_copy]() { callback_copy(); });
+        if (callback_thread_.joinable())
+        {
+          callback_thread_.join();
+        }
+        VDA5050_INFO(
+          "[" + id_ + "] Heartbeat monitoring stopped after timeout");
+        return;
```
sauk2: Do you think the callback from be invoked directly from the current thread instead of a new thread? Because the callback thread is immediately joined so there in no concurrency benefit

vda5050_master/include/vda5050_master/communication/heartbeat.hpp
```diff
@@ -0,0 +1,249 @@
...
+      state_ = HeartbeatState::STOPPED;
+    }
+
+    VDA5050_INFO("Stopped Connection heartbeat listener");
+  }
+
+  virtual std::chrono::steady_clock::time_point get_current_time()
+  {
+    return std::chrono::steady_clock::now();
+  }
```
sauk2: Change to `system_clock`

vda5050_master/include/vda5050_master/communication/heartbeat.hpp
```diff
@@ -0,0 +1,249 @@
...
+    int time_since_last_connection_report;
+    {
+      std::lock_guard<std::mutex> lock(last_connection_report_mutex_);
+      time_since_last_connection_report =
+        std::chrono::duration_cast<std::chrono::seconds>(
+          current_time - last_connection_report_)
+          .count();
+    }
+
+    if (std::abs(time_since_last_connection_report) > heartbeat_interval_)
```
sauk2: I think there may be no need to use `abs()` here as `steady_clock` is always supposed to go forward

vda5050_master/src/vda5050_master/master.cpp
```diff
@@ -0,0 +1,455 @@
...
+
+  std::lock_guard<std::mutex> lock(agv_mutex_);
+
+  if (!get_agv_by_id(agv_id))
+  {
+    VDA5050_WARN("[VDA5050Master] Cannot offboard: AGV not found: {}", agv_id);
+    return;
+  }
+
+  agvs_.erase(agv_id);
```
sauk2: Do you think it's a better idea to call `stop()` first before destroying the AGV object?

sauk2 (approved):
Thanks for the changes! I have a few final comments regarding threading behaviour and some possible race conditions.

Other than that, this one looks good to be merged!

vda5050_master/src/agv/agv.cpp
tanjpg: This part is to guard against 
1. The conditional variable checks that stop_processing is false ( In method `process_queues`)
2. `stop_processing_ = true;` ( In method `stop_queue_processor`)
3. `queue_cv_.notify_one();`  is called ( In method `stop_queue_processor`)
4. Then `queue_cv_.wait(....` is called ( In method `process_queues`)

Which will cause indefinite wait since the notification call is already done.

tanjpg: en.cppreference.com/w/cpp/thread/condition_variable/wait

```cpp
while (!pred())
    wait(lock);.
```

The race condition, i believe may happen between predicate check and `wait(lock)`

vda5050_master/include/vda5050_master/communication/heartbeat.hpp
tanjpg: I will be converting all the time related variables to use `system_clock` instead, as mentioned [here](https://github.com/ros-industrial/vda5050_client/pull/20#discussion_r2710698626), so we would still need to retain this to trigger timeout in the event of clock discontinuities.

vda5050_master/src/agv/agv.cpp
tanjpg: yes, missed this. will add a mutex to guard access to `state_heartbeat_`

vda5050_master/include/vda5050_master/agv/agv.hpp
tanjpg: Fixed in [this commit](github.com/ros-industrial/vda5050_client/pull/20/commits/f396b34439161c9e2f9c8f05ed800a7405176097)

vda5050_master/include/vda5050_master/communication/heartbeat.hpp
tanjpg: fixed in [this commit](github.com/ros-industrial/vda5050_client/pull/20/commits/f396b34439161c9e2f9c8f05ed800a7405176097)

vda5050_master/include/vda5050_master/communication/heartbeat.hpp
tanjpg: [Commit in question](github.com/ros-industrial/vda5050_client/pull/20/commits/f396b34439161c9e2f9c8f05ed800a7405176097)

vda5050_master/src/vda5050_master/master.cpp
tanjpg: [Added stop before erasing](https://github.com/ros-industrial/vda5050_client/pull/20/changes/58fa5c4860a4d445870b3bd712b60bc5ced16f7e)

Also added a separate commit [here](https://github.com/ros-industrial/vda5050_client/pull/20/changes/b347aaa1efffbee388739badd7a304a6ba03bc99) that `move()`s the offboarded AGV and release the mutex to allow for longer running `stop()` actions.

vda5050_master/src/agv/agv.cpp
tanjpg: Added the mutex for `state_heartbeat_` [here](https://github.com/ros-industrial/vda5050_client/pull/20/changes/34d19c04d1e9db4e87e05cb695d025a0de707e83)

vda5050_master/src/agv/agv.cpp
tanjpg: This was missed out in the rework to add `start` and `pause`

[This commit]() rectifies the required behavior for stop processing.

- Now the stop condition is either `stop_processing_` is `true`, **OR** both queues are empty.

Additionally, [tests have been added](https://github.com/tanjpg/vda5050_client/blob/f0913953c39d6f936799e0c07556b5e3eba2d8ad/vda5050_master/test/agv/agv_lifecycle_test.cpp#L353) to validate the following behavior (which could have caught this)
- `stop()` breaks the ` process_queues()` loop, and clears `order_queue` and `instant_actions_queue`
- `pause()` breaks the ` process_queues()` loop, but **does not** clear `order_queue` and `instant_actions_queue`

vda5050_master/src/agv/agv.cpp
tanjpg: In [this commit](https://github.com/ros-industrial/vda5050_client/pull/20/changes/4e3e04cffd2f18af30b2ae7039bcdaf5227f8b7a)

1. Added a new `thread_mutex_`
2. `stop_processing_` is now handled by `queue_mutex` so it doesnt need to be atomic (same for `queue_processor_running_` being handled by `thread_mutex`)
3. `start_queue_processor()` is now protected by `thread_mutex_` and all `queue_processor_running_` operations and checks are also protected by such.
4. I have also additionally made a small fix to free up the new `thread_mutex` during `stop_queue_processor()` earlier by `std::move`ing the remaining running thread to `join()` without holding the mutex.

vda5050_master/include/vda5050_master/communication/heartbeat.hpp
tanjpg: Technically I do this check in [`setup_heartbeat()`](https://github.com/tanjpg/vda5050_client/blob/ci-ready/vda-master/vda5050_master/src/agv/agv.cpp#L244) when this is called via the `AGV` object's method

```cpp
void AGV::setup_heartbeat()
{
  std::lock_guard<std::mutex> lock(heartbeat_mutex_);

  if (state_heartbeat_)
  {
    return;  // Already set up
  }
...
```

However, I have additionally added checks in [this commit](https://github.com/ros-industrial/vda5050_client/pull/20/changes/66e8ab726edfbc90541b3a8429c5b1ebf04e479e), as it is still a bug nonetheless

vda5050_master/include/vda5050_master/communication/heartbeat.hpp
tanjpg: Removed [here](github.com/ros-industrial/vda5050_client/pull/20/commits/51c1250b5c1c2850ff592f782c69781ea3b83e24)

sauk2 (commented):
Thanks for quick fixes! Waiting for @Briancbn before merging this in 🚀

Briancbn (changes requested):
@tanjpg @sauk2 Good job guys! Some last sanity comments from me
- `VDA5050Master` and `AGV` class should not be in global namespace (is this intentional?)
- `heartbeat.hpp` header guard did not cover the `#include`.
- Function documentation format should align with the existing repo style `\brief`, `\param` and `\return`.
- `Heartbeat` Class destructor should be `virtual`. 
- Implementation of `Heartbeat` class should be inside `.cpp` file instead.

vda5050_master/include/vda5050_master/vda5050_master/master.hpp
```diff
@@ -0,0 +1,318 @@
...
+  // ============================================================================
+
+  /**
+   * @brief Get a shared pointer to an onboarded AGV
+   * @param manufacturer Manufacturer name
+   * @param serial_number Serial number
+   * @return Shared pointer to AGV, or nullptr if not onboarded
+   */
+  std::shared_ptr<AGV> get_agv(
+    const std::string& manufacturer, const std::string& serial_number) const;
```
Briancbn, line 158: Should this return `const` instead?
```cpp
std::shared_ptr<const AGV> get_agv(
```

tanjpg:
@Briancbn 

> `VDA5050Master` and `AGV` class should not be in global namespace (is this intentional?)

Do you have a preferred namespace? `vda5050_master`?

vda5050_master/include/vda5050_master/vda5050_master/master.hpp
Briancbn: Sorry my bad, upon further understanding I realized the non-const function of `AGV` is meant to be called outside of the master class. I will close this.

tanjpg:
@Briancbn 

>     * `VDA5050Master` and `AGV` class should not be in global namespace (is this intentional?)

Added both to `vda5050_master` namespace in [this commit](https://github.com/tanjpg/vda5050_client/commit/723af3a57cd37843266d3486d8526691cbbeb01a)

>     * `heartbeat.hpp` header guard did not cover the `#include`.

Shifted [here](https://github.com/ros-industrial/vda5050_client/pull/20/changes/70807b33715f1728bab0be978b9072f85dfec504#diff-b4ff928741a2da71434ca69ee946d2f9c57f027979822ca1564ddfc5ce7bb5b8R19-R20)

>     * `Heartbeat` Class destructor should be `virtual`.
Added [here](https://github.com/ros-industrial/vda5050_client/pull/20/changes/70807b33715f1728bab0be978b9072f85dfec504#diff-b4ff928741a2da71434ca69ee946d2f9c57f027979822ca1564ddfc5ce7bb5b8R55)

>     * Implementation of `Heartbeat` class should be inside `.cpp` file instead.

Fixed in [This commit](https://github.com/tanjpg/vda5050_client/commit/70807b33715f1728bab0be978b9072f85dfec504)

>     *Function documentation format should align with the existing repo style \brief, \param and \return.

Apologies on this, will change this in a separate PR

Briancbn (approved):
lgtm!

tanjpg:
Thanks all, will keep PRs smaller subsequently

PR #22, opened 2026-01-08
feat(json validators): add JSON validation functions

shawnkchan:
## Summary
This PR adds functions to enable JSON validation logic using the nlohmann-json-schema-validator library.

### DIFF
--- a/vda5050_json_utils/CMakeLists.txt
+++ b/vda5050_json_utils/CMakeLists.txt
@@ -8,7 +8,7 @@ endif()
 find_package(ament_cmake REQUIRED)
 find_package(vda5050_types REQUIRED)
 find_package(nlohmann_json REQUIRED)
-find_package(vda5050_msgs REQUIRED)
+find_package(vda5050_interfaces REQUIRED)
 
 include(GNUInstallDirs)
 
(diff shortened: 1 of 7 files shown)


shawnkchan:
@sauk2 raising this draft PR to get your feedback on my validation logic in `validators.hpp`. 

Currently I'm only testing json validation for the connection schema. If all is good, I'll write the remaining tests for the other schemas as they should be similar to the connection schema generation and unit test.

vda5050_json_utils/include/vda5050_json_utils/validators.hpp
```diff
@@ -0,0 +1,130 @@
...
+  nlohmann::json_schema::json_validator validator(
+    nullptr, date_time_format_checker);
+
+  try
+  {
+    validator.set_root_schema(schema);
+  }
+  catch (const std::exception& e)
+  {
+    std::cerr << "Validation of schema failed: " << e.what() << "\n";
```
shawnkchan, line 114: @sauk2 This is a placeholder, I assume that we'd want to output the VDA5050 errors instead here?

vda5050_json_utils/CMakeLists.txt
```diff
@@ -8,15 +8,15 @@ endif()
...
+find_package(vda5050_interfaces REQUIRED)
 
 include(GNUInstallDirs)
 
 option(ENABLE_ROS2 "Enable ROS 2 message support in JSON utilities" OFF)
 
 if(ENABLE_ROS2)
   message(STATUS "Building vda5050_json_utils with ROS 2 support")
-  find_package(vda5050_msgs REQUIRED)
+  find_package(vda5050_interfaces REQUIRED)
```
shawnkchan, line 19: @sauk2 I'm using the updated vda5050_interfaces library (previously I still had it named as vda5050_msgs so I had no issues building the project), but I think referencing to vda5050_interfaces instead of vda5050_msgs is now causing the CI to fail

PR #25, opened 2026-01-13
Feat (CI) : Run Mosquitto in workflows to spin up local 

tanjpg:
As mentioned by @sauk2 , the `test.mosquitto.org:1883` server is relatively flaky and is not advised to be used for public testing of infrastructure. to prevent constant CI failures from intermittent downtime of the server, this PR attempts to add in a step to run mosquitto locally for testing.


sauk2 (approved):
Thanks for this! It was a much needed change.

PR #27, opened 2026-01-15
feat(humble): add support for humble

### LINKED ISSUE #26: Build failure for humble
The build currently is failing on humble due to missing `#include <atomic>` for test, I think it might be a gcc/clang version issue. I am using gcc 11 shipped with ubuntu 22.04.

```
.../vda5050_ws/src/vda5050_client/vda5050_core/test/integration/mqtt_client/test_paho_mqtt_client.cpp:34:8: error: ‘atomic_int’ is not a member of ‘std’
   34 |   std::atomic_int message_count{0};
```

We might want to add a pipeline for humble.

Briancbn:
This PR fixes the build issues for gcc-11, clang-14, 22.04 and humble.

Key changes
- Enable humble build as matrix strategy in GHA, this should make adding new distros easier
- resolve #26 ([failed jobs](https://github.com/Briancbn/vda5050_client/actions/runs/20983794175))
- add c++ 17 to address `std::optional` not found from clang 14 ([failed jobs](https://github.com/Briancbn/vda5050_client/actions/runs/21023979259))
- Only enable linters at their minimal supported versions, as they only ships with 24.04 and jazzy.

### DIFF
--- a/vda5050_core/CMakeLists.txt
+++ b/vda5050_core/CMakeLists.txt
@@ -93,13 +93,33 @@ if(BUILD_TESTING)
     FILTERS "${cpplint_filters}"
   )
 
-  find_package(ament_cmake_clang_format REQUIRED)
-  ament_clang_format(
-    CONFIG_FILE "${CMAKE_CURRENT_SOURCE_DIR}/../.clang-format"
-  )
+  # Only run ament_clang_format for 16.0.0 and higher
+  find_program(CLANG_FORMAT_EXECUTABLE NAMES clang-format)
+  if(CLANG_FORMAT_EXECUTABLE)
+    execute_process(
+        COMMAND ${CLANG_FORMAT_EXECUTABLE} --version
+        OUTPUT_VARIABLE CLANG_FORMAT_VERSION_OUTPUT
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
 
-  find_package(ament_cmake_copyright REQUIRED)
-  ament_copyright()
+    # Parse the version number from the output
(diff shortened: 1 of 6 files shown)


vda5050_core/CMakeLists.txt
```diff
@@ -93,13 +97,33 @@ if(BUILD_TESTING)
...
+        CONFIG_FILE "${CMAKE_CURRENT_SOURCE_DIR}/../.clang-format"
+      )
+    endif()
+  endif()
+
+  # Disable copyright check for older ament_copyright versions
+  find_package(ament_cmake_copyright 0.13.3 QUIET)
+  if(ament_cmake_copyright_FOUND)
+    ament_copyright()
+  endif()
```
sauk2: This code will have to be added for each package. There are already 2-3 new packages that will be added to this repo that will have to copy this over. Is it instead possible to create a `.cmake` file that can either be in the root of the repo or somewhere else that is more appropriate? Each package can then include this CMake module file during test generation.

sauk2 (commented):
Thanks for the addition of a `humble` job! I don't see any changes required to the code. Just on concern regarding usability

vda5050_core/CMakeLists.txt
Briancbn: I can create a `cmake` folder in the project root with a `vda5050_lint_common.cmake` file containing a macro under the same name with all the linters. Let me know your thoughts.

sauk2 (approved):
Thanks for the changes. Merging this in :rocket:

Issue #21, opened 2026-01-06
Add an container to store callbacks when connect is not called


sauk2:
## Summary

In the current implementation of the Paho MQTT client, the `subscribe(...)` function does not perform a subscription if `connect()` is not called before it. So I was thinking we could add a map to store callbacks when subscription requests are invoked before connecting. The `connect()` function will check this container and perform subscriptions upon invocation.

Briancbn:
I am not entirely clear on the use case for this. It seems to me that "pub/sub before connect" is a behavior should not be allowed. Is this related to handling intermittent connection?

sauk2:
It was due to the fact that we are not returning anything from the `subscribe(...)` method so a user won't know if the subscription was successful or not. Maybe adding a container to store callbacks is overkill, instead I think we can change the return type of this function from void to something else.

PR #28, opened 2026-01-19
feat(execution): add execution utilities

sauk2:
## Summary

This PR adds an initial set of classes for building the execution system. More information can be found in the design doc under the `docs/` folder. Tests have been added for each class along with a `README` with examples inside `vda5050_execution` package.


vda5050_execution/include/vda5050_execution/callback_registry.hpp
```diff
@@ -0,0 +1,81 @@
...
+      cb(std::static_pointer_cast<EventT>(event));
+    };
+
+    callbacks_[std::type_index(typeid(EventT))].push_back(std::move(wrapper));
+  }
+
+  void dispatch(std::shared_ptr<EventBase> event) const;
+
+  template <typename UpdateT>
+  void register_provider(std::function<void(std::shared_ptr<UpdateT>)> provider)
```
Briancbn: I suggest to make `providers` its own registry. The `provider` doesn't interacts with `callback` and the `Provider` class is basically a wrapper of the function here. The `static_assert` is also called repeatedly for the same check.

vda5050_execution/include/vda5050_execution/event_queue.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+#include "vda5050_execution/event.hpp"
+
+namespace vda5050_execution {
+
+class EventQueue
+{
+public:
+  template <typename EventT, typename... Args>
+  void push(Args&&... args)
```
Briancbn: Is this class intended to be thread-safe? If so, all read-write operation to the `std::queue` should lock the mutex.

vda5050_execution/include/vda5050_execution/provider.hpp
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  template <typename UpdateT>
+  void on(std::function<void(std::shared_ptr<UpdateT>)> callback)
+  {
+    callback_registry_.register_provider<UpdateT>(std::move(callback));
+  }
+
+private:
+  CallbackRegistry callback_registry_;
```
Briancbn: I suggest to make Provider its own registry as suggested above.

vda5050_execution/src/vda5050_execution/callback_registry.cpp
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+#include "vda5050_execution/callback_registry.hpp"
+
+namespace vda5050_execution {
+
+//=============================================================================
+void CallbackRegistry::dispatch(std::shared_ptr<EventBase> event) const
+{
+  auto it = callbacks_.find(event->get_type());
```
Briancbn: Just a reminder,`event` might be `nullptr`, but  it's up to you whether you want to handle the exception here.

vda5050_execution/src/vda5050_execution/event_queue.cpp
```diff
@@ -0,0 +1,42 @@
...
+  std::lock_guard<std::mutex> lock(mutex_);
+  return queue_.empty();
+}
+
+//=============================================================================
+std::shared_ptr<EventBase> EventQueue::pop()
+{
+  if (queue_.empty()) return nullptr;
+
+  std::lock_guard<std::mutex> lock(mutex_);
```
Briancbn: Checking queue empty should also be under mutex lock?

```suggestion
  std::lock_guard<std::mutex> lock(mutex_);

  if (queue_.empty()) return nullptr;
```

vda5050_execution/src/vda5050_execution/execution_engine.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+namespace vda5050_execution {
+
+void ExecutionEngine::step()
+{
+  if (!event_queue_.empty())
+  {
+    std::shared_ptr<EventBase> event = event_queue_.pop();
+    callback_registry_.dispatch(event);
+  }
```
Briancbn: I suggest to check if the `pop()` returned `event` is a `nullptr` instead, as the `pop()` method already check queue `empty()` inside.

Potential race condition:
Between the `if` condition and `pop` here, event could potentially be removed from another thread causing event to be a `nullptr`.

Briancbn (changes requested):
Thanks @sauk2, good work on the execution engine and provider. I have some comments.

I am not that familiar with the vda5050 topic to type mapping, @tanjpg could you take a look at the `ProtocolAdapter`? I am ok with the design and implementation of the `ProtocolAdapter`, but I think the error handling for `ProtocolAdapter` still need some work, this could be the result of https://github.com/ros-industrial/vda5050_client/issues/21#issuecomment-3775944610 and can be addressed at a later time.

vda5050_execution/src/vda5050_execution/protocol_adapter.cpp
```diff
@@ -0,0 +1,71 @@
...
+  const std::string& interface, const std::string& version,
+  const std::string& manufacturer, const std::string serial_number)
+: mqtt_client_(mqtt_client),
+  interface_(interface),
+  version_(version),
+  manufacturer_(manufacturer),
+  serial_number_(serial_number)
+{
+  std::string topic_prefix = fmt::format(
+    "{}/{}/{}/{}", interface_, version_, manufacturer, serial_number_);
```
tanjpg: @sauk2 small question, is it intended to be `manufacturer` instead of `manufacturer_`?  just visual consistency

tanjpg:
> Thanks @sauk2, good work on the execution engine and provider. I have some comments.
> 
> I am not that familiar with the vda5050 topic to type mapping, @tanjpg could you take a look at the `ProtocolAdapter`? I am ok with the design and implementation of the `ProtocolAdapter`, but I think the error handling for `ProtocolAdapter` still need some work, this could be the result of [#21 (comment)](https://github.com/ros-industrial/vda5050_client/issues/21#issuecomment-3775944610) and can be addressed at a later time.


The topic mapping is as per the vda5050 convention

vda5050_execution/include/vda5050_execution/callback_registry.hpp
sauk2: I am also thinking if there is a need to have a separate `CallbackRegistry` class for `EventBase`. Because I am forwarding arguments there too. I am thinking of transferring all that code into `ExecutionEngine`

vda5050_execution/src/vda5050_execution/callback_registry.cpp
sauk2: Building on to what I said about removing `CallbackRegistry` altogether, it will also help with this. Just check for `nullptr` once.

vda5050_execution/include/vda5050_execution/callback_registry.hpp
sauk2: > I suggest to make `providers` its own registry. The `provider` doesn't interacts with `callback` and the `Provider` class is basically a wrapper of the function here. The `static_assert` is also called repeatedly for the same check.

Addressed in f45773397efd36184e73d9ba0d63ea7174d4063f

vda5050_execution/include/vda5050_execution/event_queue.hpp
sauk2: Addressed in f193a9fca3825b74d289cb6e8d6438bc6c429047

vda5050_execution/src/vda5050_execution/callback_registry.cpp
sauk2: Addressed in feba6be35c81ff5b26bb2bbbacada02c09fb2b2b

vda5050_execution/src/vda5050_execution/protocol_adapter.cpp
sauk2: Addressed in 2bb849c9e448cb0f6c335beab37b410febd4404e

.clang-format
```diff
@@ -12,10 +12,11 @@ BraceWrapping:
   AfterControlStatement: Always
   AfterFunction: true
   AfterNamespace: false
-  AfterStruct: true
+  AfterStruct:  true
```
Briancbn: extra space

vda5050_execution/include/vda5050_execution/callback_registry.hpp
Briancbn: > I am also thinking if there is a need to have a separate `CallbackRegistry` class for `EventBase`. Because I am forwarding arguments there too. I am thinking of transferring all that code into `ExecutionEngine`

I agree. Please proceed. Also don't forget to also remove the section on the callback registry in the doc.

.clang-format
sauk2: Addressed in 3ffff1bbc2566bf3175ac3594410664d4344fa21

sauk2:
@Briancbn Ready for your review!

Briancbn (approved):

PR #31, opened 2026-01-23
feat: order validator changes

sauk2:
## Summary

This PR makes changes to the validation functions in the [`feat/dev-orderGraphValidator`](https://github.com/ros-industrial/vda5050_client/tree/feat/dev-orderGraphValidator) branch.


vda5050_core/src/vda5050_core/client/order/order_graph_validator.cpp
```diff
@@ -0,0 +1,238 @@
...
+namespace vda5050_core {
+
+namespace order {
+
+//=============================================================================
+ValidationResult is_valid_graph(const vda5050_types::Order& order)
+{
+  ValidationResult res;
+
+  auto add_error = [&](
```
tanjpg, line 34: @sauk2 is the intention of this method to flag out all the errors found in the order? I see that errors are appended as the method is executed, rather than returned at the first error that it finds .

If the intention is to log out all errors, might need to see if there is a need to identify dependant errors that may be propagated from one error. For e.g:

1. If a Node Sequence ID is Odd (Wrong)
2. If an Edge Sequence ID is +1 increment from Node Sequence ID (Correct)
- This inherently adds a new error that Edge Sequence ID is wrong because it will be Even (despite not checking `Edge Sequence ID %2 == 0`)

Edit: I realized that there are some checks that will lead to return-ing of the method. I suppose if the error does not break the other checks then it will continue

vda5050_core/src/vda5050_core/client/order/order_graph_validator.cpp
```diff
@@ -0,0 +1,238 @@
...
+      {
+        add_error(
+          "Edge connectivity mismatch.", {{errors::RefEdgeId, edge.edge_id},
+                                          {errors::RefNodeId, node.node_id}});
+      }
+
+      // Check for proper base and horizon separation
+      if (!edge.released)
+      {
+        horizon_reached = true;
```
tanjpg, line 153: @sauk2 I think there may be a check that is not fully captured here.

`An edge can only be released if both start and end nodes are released`

This would mean that the first element in a horizon should always be an edge. So there seems to be a missing check to ensure that the thing that triggers `horizon_reached = true` shouldnt be because `!node.released`

vda5050_core/src/vda5050_core/client/order/order_graph_validator.cpp
```diff
@@ -0,0 +1,238 @@
...
+      next_order.order_update_id != 0 ||
+      next_order.nodes.front().sequence_id != 0)
+    {
+      add_error(
+        "New order must start with update 0 and sequence 0.",
+        {{errors::RefSequenceId,
+          std::to_string(next_order.nodes.front().sequence_id)}});
+      return res;
+    }
+  }
```
tanjpg: @sauk2 Is this inherently already checked in `is_valid_graph`? Regardless of whether or not the order_id is new or the base order, the Order Update ID, if it is 0, it should always start with Sequence ID Node 0

I think the check on whether an order is an update or not can be done outside this method, and only call this if the `order_id` is different

vda5050_core/src/vda5050_core/client/order/order_graph_validator.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+  // Check if a stitching node exists
+  if (stitching_it == next_order.nodes.end())
+  {
+    add_error("Last released node not found in new message.", {});
+    return res;
+  }
+
+  return res;
+}
```
tanjpg, line 217: @sauk2 Based on the image below, i think this part just need to check that the last node of the base is identical to the first node of the update order. If that is not true, then it immediately is not valid for stitching.

Side note, I wonder if overloading `==` for all vda5050 custom object definitions would be useful for these kind of methods and checks.

Image

vda5050_core/src/vda5050_core/client/order/order_graph_validator.cpp
sauk2: Thanks for catching this! Instead of changing the `horizon_reached` flag after finding an unreleased node, now I am changing the flag when I find the first unreleased edge followed by making sure the successive nodes are unreleased.

Addressed in commit 4822151366261ec8f599bcb7bc7ad9b16285855b

vda5050_core/src/vda5050_core/client/order/order_graph_validator.cpp
sauk2: You're right.! I removed this check

Also, I have restructured the checks to be this way: check if the order update is increasing -> check validity of the order graph -> check if the graphs can be stitched. 

Addressed in commit 4822151366261ec8f599bcb7bc7ad9b16285855b

vda5050_core/src/vda5050_core/client/order/order_graph_validator.cpp
tanjpg: LGTM!

vda5050_core/src/vda5050_core/client/order/order_graph_validator.cpp
sauk2: Yes, the intention is to provide as much context as possible to the MC in a single pass. In the example you gave, the edge having an even sequence is a consequence of the node having an odd sequence. Strictly adhering to the specification, both are distinct violations (node must be even, edge must be odd).

The current design prioritizes spec-based checking to ensure the user sees the full extent of the malformed data. However, if you think it would be better to add some awareness to reduce the amount of errors we generate, we could implement preconditions, for your example, only checking the edge sequence if the sequence increment from the node was valid. This would suppress the secondary error (edge is even) while still flagging the root cause (node is odd).

Let me know your thoughts!

vda5050_core/src/vda5050_core/client/order/order_graph_validator.cpp
sauk2: The reason I am finding the matching node in the update rather than just checking the first node is to accommodate MCs that send overlapping released nodes. This ensures that even if the incoming message contains redundant nodes already present in the previous base, the validation succeeds as long as the 'stitching point' (the last released node) is present and identical.

> Side note, I wonder if overloading == for all vda5050 custom object definitions would be useful for these kind of methods and checks.

Yeah that's a great idea! The `==` operators are already overloaded in all `vda5050_types` structs, I just changed to direct equating instead of individually checking in this commit d4014a4b4b58a99a1a6d33a1f0e4ca363dc0151c

tanjpg: Hm im not too sure if i understand what you mean. From V2.0.0:
```bash
Also note that node 7 is the only base node that is sent again. Since the base cannot be changed, a retransmission of nodes 6 and 4 is not valid.
```

How I interpret this is that there must be **only** 1 overlapping node (which is stitching node). Anything more than that is not valid (even if it is redundant). Let me know if you interpret this differently!

An alternative is that we accommodate this but manually remove those nodes if we detect overlaps. however if this order validator is to adhere to the vda5050 standard then maybe this is not ideal. 

Let me know your thoughts!

vda5050_core/src/vda5050_core/client/order/order_graph_validator.cpp
tanjpg: We can keep this current format, i think its fine. thanks for the work here!

tanjpg (changes requested):
Just the last part on overlapping base nodes to align.

vda5050_core/src/vda5050_core/client/order/order_graph_validator.cpp
sauk2: Makes sense. 

In one of our previous implementations, we used to allow overlapping nodes and blindly ignore all the nodes before the stitching point in the incoming order update. But maybe it's better to strictly align to the spec and not accept any order update that contains more base nodes than necessary.

Will change to reading the only first node of the update to stitch.

sauk2:
> Just the last part on overlapping base nodes to align.

Thanks for the feedback! I have made the agreed upon changes

tanjpg:
LGTM! Thanks for the changes

tanjpg (approved):

PR #34, opened 2026-01-30
ci(gha): add asan and tsan for jazzy

Briancbn:
This PR enable ASAN and TSAN for jazzy.

It fixes the mqtt thread sanitizer error mentioned in https://github.com/eclipse-paho/paho.mqtt.cpp/issues/481 by bumping the version used in ci to `1.5.0`. 
I don't think this is needed for non-developers.

PR #35, opened 2026-02-04
Added fixes to sanitizer test issues 

tanjpg:
This pull requests addresses issues faced when including TSAN/ASAN for sanitization tests in the CI

Note that I have [added in the minor change](https://github.com/tanjpg/vda5050_client/commit/2159ae3e7d8157306c851c7db0c811599fbdc331) made that was meant to be in [this PR](https://github.com/ros-industrial/vda5050_client/pull/34), so @Briancbn seek your advice if I should wait for your PR first before addressing the rest. 

The following changes were made to Address the failing sanitizer CI:

## 0. Potential missing flags in the CI

I think you [need the `-fno-omit-frame-pointer` ](https://github.com/tanjpg/vda5050_client/commit/8356e425b1c725de204d3a414998324b30906006)for the ASAN configuration as well. Not too sure exactly what is going on under the hood, but from testing it seems to be able to prevent the ASAN CI pipeline from failing. 

## 1. Using of `gmtime_r()' instead of gmtime() and localtime_r() instead of localtime()

As mentioned [here](https://en.cppreference.com/w/c/chrono/gmtime.html) it seems like `gmtime()` is not thread safe due to returning of a static shared buffer. [Same with localtime()](https://en.cppreference.com/w/c/chrono/localtime.html)

[This commit](https://github.com/tanjpg/vda5050_client/commit/8c5c69b3116b66fda498f6631bca65f4c13aa005) thus instead use the threadsafe `localtime_r()` and  `gmtime_r()` function that takes in a `std::tm` buffer as well

```cpp
std::tm localtime;
localtime_r(&now_t, &localtime);
```

```cpp
std::tm tm_buf;
gmtime_r(&time_sec, &tm_buf);
.....
oss << std::put_time(&tm_buf, vda5050_types::ISO8601_FORMAT);
```

## 2. Adding anonymous namespaces for `EventA` and `EventB`
 
There seems to be two `EventA` classes and `EventB` classes in [test_event_queue.cpp](https://github.com/ros-industrial/vda5050_client/blob/develop/vda5050_execution/test/unit/test_event_queue.cpp#L27) and [test_execution_engine.cpp](https://github.com/ros-industrial/vda5050_client/blob/develop/vda5050_execution/test/unit/test_execution_engine.cpp#L27), that i think may be causing a potential ODR violation.

I [added a simple anonymous namespace](https://github.com/tanjpg/vda5050_client/commit/0e44c0df2161a33ab9cd54ee44fa86a5c90a4144) to address this.

## 3. Issues with Mockheartbeat class (Used only in unit tests)

I [made the `MockHeartbeatListener` class atomic](https://github.com/tanjpg/vda5050_client/commit/142088693f9196984c65a59959701b1b5b72433c) to ensure no data races when `.get()` and `.load()` is being used in different threads

## 4. Mutex issues with `vda5050_master::AGV`

[Fixed the potential threading issues](https://github.com/tanjpg/vda5050_client/commit/f461331b5f0a0902b7339e334dc70bd0dbb960cb) where other threads may call the `HeartbeatListener` in the `vda5050_master::AGV` object when the `HeartbeatListener`'s `stop_connection_heartbeat()` method is triggered using mutexes and transfer of pointer ownership with `std::move`

### DIFF
--- a/vda5050_master/test/integration/heartbeat.cpp
+++ b/vda5050_master/test/integration/heartbeat.cpp
@@ -217,7 +217,8 @@ TEST(HeartbeatListenerTest, StateIsRunningWhileCallbackExecutes)
   std::atomic_bool was_running_during_callback{false};
 
   // We need a raw pointer to check get_state() from within callback
-  MockHeartbeatListener* listener_ptr = nullptr;
+  // Must be atomic because the callback thread may read it while main thread writes
+  std::atomic<MockHeartbeatListener*> listener_ptr{nullptr};
 
   auto hb_listener = std::make_unique<MockHeartbeatListener>(
     "test_listener", vda5050_master::ConnectionHeartbeatInterval,
--- a/vda5050_master/test/integration/heartbeat.cpp
+++ b/vda5050_master/test/integration/heartbeat.cpp
@@ -22,6 +22,7 @@
 
 #include <atomic>
 #include <chrono>
+#include <memory>
 #include <thread>
 #include <vector>
 
@@ -142,7 +143,7 @@ TEST(HeartbeatListenerTest, HeartbeatReceivedNoTimeout)
 TEST(HeartbeatListenerTest, HeartbeatNotReceivedTimeout)
 {
   std::atomic_bool heartbeat_failed{false};
-  auto hb_listener = MockHeartbeatListener(
+  auto hb_listener = std::make_unique<MockHeartbeatListener>(
     "test_listener", vda5050_master::ConnectionHeartbeatInterval,
     [&heartbeat_failed]() {
       // Timeout callback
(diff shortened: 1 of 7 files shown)


vda5050_master/test/integration/heartbeat.cpp
sauk2: I just noticed this line which is present in multiple tests. It is unrelated to this PR but I think this is a risky line. Since `MockHeartbeatListner` is stack allocated in this test, calling a destructor explicitly might cause a double free error. 

 Would be better if you could create this object on the heap through a `std::shared_ptr` or `std::unique_ptr` and then call a `.reset()` to invoke the destructor.

sauk2 (approved):
Thanks for the thorough investigation and fixes! It is great to finally have ASAN/TSAN enabled. I have left one minor, unrelated comment below, but the rest of the PR looks great to go.

vda5050_master/test/integration/heartbeat.cpp
tanjpg: I have made changes in the [Latest commit](https://github.com/ros-industrial/vda5050_client/pull/35/changes/27cb295a657bdd5ce0c216f914a8c7524a41deab) do take a quick look, and if it is ok, I will merge it in.

PR #36, opened 2026-03-03
feat(execution): change base classes and execution handler 

sauk2:
## Summary
 
This PR introduces an end-to-end Strategy-Context-Handler design for the execution system. Following are the change from the previous iteration

- A common `Base` class has been introduced along with 3 types of data carriers - `EventBase`, `UpdateBase` and `ResourceBase`
- The `Engine` is now part of a Strategy (derived from `StrategyInterface`) that is used to push Events (derived from `EventBase`) to the queue.
- The `Provider` is now part of the Context (derived from `ContextInterface`) and is used to received Updates (derived from `UpdateBase`). A typical implementation of the Context will contain some form of a container to store Updates and Resources. 
- Resources (derived from `ResourceBase`) are meant to be static and should not be modified mid-execution.
- A `Handler` runs the Strategy by taking in a Context at each step. It has a blocking `spin` and a non-blocking `spin_once` option.

## Testing and Examples

There are a set of examples in the `examples/` directory that can be used to understand the usage.

More information can be found in the `docs/` directory and the local `README.md`.


sauk2:
cc @DillonChew98

sauk2:
@Briancbn this is ready for your review!

vda5050_execution/include/vda5050_execution/handler.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+namespace vda5050_execution {
+
+class Handler : public std::enable_shared_from_this<Handler>
+{
+public:
+  template <typename... Strategies>
+  static std::shared_ptr<Handler> make(
+    std::shared_ptr<ContextInterface> context,
+    std::shared_ptr<Strategies>... strategies)
```
Briancbn: The variadic and template here feels unnecessary to me. Could we rely on the class polymorphism and pass in a vector here?

vda5050_execution/include/vda5050_execution/engine.hpp
```diff
@@ -56,18 +61,60 @@ class ExecutionEngine : public std::enable_shared_from_this<ExecutionEngine>
       cb(std::static_pointer_cast<EventT>(event));
     };
 
+    std::lock_guard<std::mutex> lock(registry_mutex_);
     callbacks_[std::type_index(typeid(EventT))].push_back(std::move(wrapper));
   }
 
+  template <typename UpdateT>
+  void wait_for(
```
Briancbn: It seems that `wait_for` here is to add a delay condition for emit. Just from the name of the function, it can be easily misled as a blocking function. I think some documentation is definitely needed here (as well as the critical event behavior). We can also change it to `add_delay`, `suspend` to make the intention a bit more explicit.

vda5050_execution/src/vda5050_execution/handler.cpp
```diff
@@ -0,0 +1,100 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "vda5050_execution/handler.hpp"
+
+namespace vda5050_execution {
+
+//=============================================================================
+void Handler::add_strategy(std::shared_ptr<StrategyInterface> strategy)
```
Briancbn, line 35: It seems that add strategy can add multiple instances of the same pointer, but remove strategy will remove the first instances of the same pointer. Is this intended?

vda5050_execution/src/vda5050_execution/handler.cpp
```diff
@@ -0,0 +1,100 @@
...
+//=============================================================================
+void Handler::remove_strategy(std::shared_ptr<StrategyInterface> strategy)
+{
+  std::lock_guard<std::mutex> lock(strategy_mutex_);
+  auto it = std::find(strategies_.begin(), strategies_.end(), strategy);
+  if (it != strategies_.end()) strategies_.erase(it);
+}
+
+//=============================================================================
+void Handler::wake()
```
Briancbn, line 57: `needs_processing_` doesn't seem to be used in anyway meaningful in the handler?

vda5050_execution/src/vda5050_execution/handler.cpp
```diff
@@ -0,0 +1,100 @@
...
+void Handler::spin_once()
+{
+  {
+    std::lock_guard<std::mutex> lock(sync_mutex_);
+    needs_processing_ = false;
+  }
+
+  std::vector<std::shared_ptr<StrategyInterface>> active_strategies;
+  {
+    std::lock_guard<std::mutex> lock(strategy_mutex_);
```
Briancbn: We don't need a mutex specifically for strategy, you can reuse the other mutex here.

vda5050_execution/src/vda5050_execution/handler.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+//=============================================================================
+void Handler::spin(std::chrono::milliseconds timeout)
+{
+  running_ = true;
+
+  while (running_)
+  {
+    std::unique_lock lock(sync_mutex_);
+    cv_.wait_for(lock, timeout, [&] { return needs_processing_ || !running_; });
```
Briancbn: Both `needs_processing_` and `running_` are atomic. No need to keep them as predicate, since there is no need to lock the mutex to check these values.

vda5050_execution/src/vda5050_execution/handler.cpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+//=============================================================================
+bool Handler::running()
+{
+  return running_;
+}
+
+//=============================================================================
+void Handler::stop()
```
Briancbn, line 106: You should `notify_all()` here to `wait_for`?

I believe this function should also be called in the class destructor.

Briancbn (changes requested):
Thanks for the changes, looks good! I have some questions about the `Handler` class in the following review comments.

P.S.
I usually do `_<function-name>` for private member function instead of `<function-name>_`. Could you double check the google style to see if they have any distinction between private attributes and private functions?

sauk2:
> P.S. I usually do `_<function-name>` for private member function instead of `<function-name>_`. Could you double check the google style to see if they have any distinction between private attributes and private functions?

I haven't found any explicit recommendation on this in any style guide, but I keep seeing articles about how it is safer to use trailing underscores to avoid specific naming patterns strictly set aside for the compiler and standard library.

Some excerpts,

From: https://stackoverflow.com/a/3651336
> In C++,
>
> 1. identifiers starting with an underscore, followed by a capital character
> 2. identifiers having two consecutive underscores anywhere
> 3. identifiers in the global namespace starting with an underscore
are reserved to the implementation. (More about this can be found [here](https://stackoverflow.com/q/228783/140719).) Rather than trying to remember these rules, many simply do not use identifiers starting with an underscore. That's why the trailing underscore was invented.
>
> However, C++ itself is old, and builds on 40 years of C (both of which never had a single company behind them), and has a standard library that has "grown" over several decades, rather than brought into being in a single act of creation. This makes for the existence of a lot of differing naming conventions. Trailing underscore for privates (or only for private data) is but one, many use other ones (not few among them arguing that, if you need underscores to tell private members from local variables, your code isn't clear enough).

From: https://stackoverflow.com/a/228797
> The rules (which are still unchanged in C++23):
>
> - Reserved in any scope, including for use as [implementation](https://stackoverflow.com/questions/4297933/c-implementation#4297974) macros:
>      - identifiers beginning with an underscore followed immediately by an uppercase letter
>      - identifiers containing adjacent underscores (or "double underscore")
> - Reserved in the global namespace:
>      -  identifiers beginning with an underscore
> - Also, everything in the std namespace is reserved. (You are allowed to add template specializations, though.)
> - User-defined literal suffix identifiers must start with an underscore (since C++11) and must not contain double underscore (reserved since C++23)

vda5050_execution/include/vda5050_execution/handler.hpp
sauk2: Changed as part of https://github.com/ros-industrial/vda5050_client/pull/36/commits/74564e926927aa36738c92b45ae4e6160b513262

vda5050_execution/src/vda5050_execution/handler.cpp
sauk2: Changed as part of https://github.com/ros-industrial/vda5050_client/pull/36/commits/74564e926927aa36738c92b45ae4e6160b513262

vda5050_execution/src/vda5050_execution/handler.cpp
sauk2: Changed as part of https://github.com/ros-industrial/vda5050_client/pull/36/commits/74564e926927aa36738c92b45ae4e6160b513262

vda5050_execution/src/vda5050_execution/handler.cpp
sauk2: Changed as part of https://github.com/ros-industrial/vda5050_client/pull/36/commits/74564e926927aa36738c92b45ae4e6160b513262

vda5050_execution/src/vda5050_execution/handler.cpp
sauk2: Changed as part of https://github.com/ros-industrial/vda5050_client/pull/36/commits/74564e926927aa36738c92b45ae4e6160b513262

vda5050_execution/src/vda5050_execution/handler.cpp
sauk2: Changed as part of https://github.com/ros-industrial/vda5050_client/pull/36/commits/03b81f86a3600f32399c417b182fa5c79bd67b80

vda5050_execution/include/vda5050_execution/engine.hpp
sauk2: Changed as part of https://github.com/ros-industrial/vda5050_client/pull/36/commits/dc3b365967c8678472223563192b07e46485a2d9

Briancbn: I suggest to use `suspend_for`, instead of `suspend_till`. "until" is more of a time point than a duration.

vda5050_execution/include/vda5050_execution/handler.hpp
Briancbn: Could you use `const &` here instead?

```c++

make(
...
const std::vector<std::shared_ptr<StrategyInterface>> & strategies = {}
)
```

sauk2: https://github.com/ros-industrial/vda5050_client/pull/36/commits/d726deb541688c45ec421ff4ee3204652af6af0c

vda5050_execution/include/vda5050_execution/engine.hpp
sauk2: https://github.com/ros-industrial/vda5050_client/pull/36/commits/6e67d6c97fa185a9f877ab05b03f32a4a29c11a5

sauk2:
@Briancbn CI is all green! I have addressed all your comments including removing trailing underscores from private functions.

Ready to be merged :rocket:

Briancbn (approved):

PR #33, opened 2026-01-29
Feat (VDA5050 master) : Using`vda5050_execution::ProtocolAdapter` to manage Pub/Sub

tanjpg:
This PR changes the method in which the `VDA5050Master` Object and its' stored `AGV` objects handles MQTT communication with the newly added `ProtocolAdapter` by @sauk2 

Main Changes:

- There will only be 1 `vda5050_core::mqtt_client` , held by `VDA5050Master` , that will [assign `ProtocolAdapter`s to individual AGVs during onboarding](https://github.com/tanjpg/vda5050_client/blob/feat/master/vda5050-execution-integration/vda5050_master/src/vda5050_master/master.cpp#L115)
- The `VDA5050Master` will not be using wildcarded topics from now on. `ProtocolAdapter`s will be assigned to specific `AGV`s with `manufacturer` and `serialNumber`
- Subscriptions will be started in the `AGV` objects themselves. Removed from `VDA5050Master`
- Removed all [virtual callbacks that were meant for overriding by external creators of `VDA5050Master`](https://github.com/tanjpg/vda5050_client/commit/6c25f6d1f6a44543993ccc476606c3d0b3cd5b20#diff-ed4bbbf8018ce0e136468b33ef8194ad35ecbf1c85b55bbe95e5019a795be92aL189). I believe we will still need some callback mechanism insertion later on, but I will remove it now until we have a concrete plan

Future notes:
-  We need a proper cleanup for `ProtocolAdapter`. I have [left a TODO placeholder for this](https://github.com/tanjpg/vda5050_client/blob/feat/master/vda5050-execution-integration/vda5050_master/src/agv/agv.cpp#L62)
- I need time to look through the `ExecutionEngine`, `EventQueue` and the like to see how this ties in with the VDA5050 Master, if any. This will come in future PRs.

PR #45, opened 2026-04-29
Using vda5050_execution::ProtocolAdapter to manage Pub/Sub 

stevelimyh:
Builds on PR #33, cherry-picks the ProtocolAdapter migration and lands the supporting fixes and API gap closure.                                                                                                  
                                                                                                                                                                                                                     Main Changes:                                                                                                                                                                                                      
- Typed QoS levels (`enum class QosLevel)` + spec-compliant `instantActions` topic name
           
- Per-AGV `ProtocolAdapter` migration (cherry-picked from #33) — single mqtt_client on `VDA5050Master`, per-AGV typed adapter assigned during onboarding, no more wildcard subscriptions

- Restored virtual callbacks via AGV back-pointer to `VDA5050Master`

- Proper AGV destructor teardown order (stop spinning → release resources → join threads); per-AGV subscriptions unsubscribe on `offboard_agv()` to prevent callbacks firing on a destroyed AGV

- New `ProtocolAdapter::unsubscribe<T>()` in `vda5050_execution`, symmetric counterpart to `subscribe<T>()`; closes the cleanup TODO flagged in #33 

- Master-side `cleanup_agv_subscriptions` helper deleted, uses the new `unsubscribe<T>() `instead of reaching past the adapter to the underlying `MqttClient`


vda5050_master/src/agv/agv.cpp
```diff
@@ -33,39 +34,98 @@ namespace vda5050_master {
...
+    protocol_adapter_->unsubscribe<vda5050_types::Factsheet>();
+    protocol_adapter_->unsubscribe<vda5050_types::Visualization>();
+  }
+  protocol_adapter_.reset();
+
   VDA5050_INFO("[AGV] AGV instance destroyed: {}", agv_id_);
 }
 
+template <typename MsgType>
+void AGV::create_subscription(void (AGV::*agv_handler)(const MsgType&), int qos)
```
Briancbn: This part should be in the hpp. 

Another pattern that you can follow is to have a `agv_impl.hpp` file to define a function that is essentially the private function, which will be included in the `.cpp` for compilation. But this pattern is more to facilitate polymorphism. You can find this pattern quite common in the rclcpp.

So in this case it would be

`agv.hpp`

```cpp
class AGV
{
public:
   ...

private:
  template <typename MsgType>
  friend void create_subscription(AGV* agv, void (AGV::*agv_handler)(const MsgType&), int qos);
}
```

`agv_impl.hpp`

```cpp
#include "vda5050_master/agv/agv_impl.hpp"

template <typename MsgType>
void create_subscription(AGV* agv, void (AGV::*agv_handler)(const MsgType&), int qos)
{
  agv->protocol_adapter_->subscribe<MsgType>(
    ...
  )
}
```

`agv.cpp`

```cpp
#include "vda5050_master/agv/agv.hpp"
#include "vda5050_master/agv/agv_impl.hpp"

...

void AGV::setup_subscriptions() {
    create_subscription<vda5050_types::Connection>(
      this, &AGV::handle_connection, static_cast<int>(ConnectionQos));
}

...
```

You can put the `create_subscription` function inside a namespace such as `detail` or `impl` if you prefer.

vda5050_master/include/vda5050_master/agv/agv.hpp
```diff
@@ -307,6 +315,19 @@ class AGV
...
+  // ============================================================================
+
+  void setup_subscriptions();
+
+  // ============================================================================
+  // Message Handlers (from MQTT callbacks)
+  // ============================================================================
+
+  template <typename MsgType>
+  void create_subscription(void (AGV::*agv_handler)(const MsgType&), int qos);
```
Briancbn: I suggest to avoid C style callback here. I recommend to use `std::function` instead. You can pass the function using `std::bind` or as a `lambda`.

I think that using `QosLevel` here can avoid the multiple `static_cast<int>` below.

vda5050_master/include/vda5050_master/agv/agv.hpp
```diff
@@ -75,18 +75,26 @@ class AGV
...
    */
   AGV(
+    std::shared_ptr<vda5050_execution::ProtocolAdapter> protocol_adapter,
     const std::string& manufacturer, const std::string& serial_number,
-    const std::string& broker_address, size_t max_queue_size = 10,
-    bool drop_oldest = true,
-    int state_heartbeat_interval = StateHeartbeatInterval);
+    size_t max_queue_size = 10, bool drop_oldest = true,
+    int state_heartbeat_interval = StateHeartbeatInterval,
+    VDA5050Master* parent = nullptr);
```
Briancbn: Maybe a `weak_ptr` is better here? It might introduce a lot more boilerplate code, let me know your thoughts here?

vda5050_master/include/vda5050_master/agv/agv.hpp
```diff
@@ -61,7 +61,7 @@ enum class AGVState
  * - Outgoing message queue for orders and instant actions
  *
  * The VDA5050Master routes incoming messages to AGV instances and the AGV
- * handles outgoing messages via transient MQTT clients.
+ * handles ingoing/outgoing messages via the VDA5050Execution ProtocolAdapter.
```
Briancbn:
```suggestion
 * handles incoming/outgoing messages via the VDA5050Execution ProtocolAdapter.
```

Briancbn (changes requested):
Thanks @stevelimyh for follow up on the previous PR. Thanks for adding the additional comments for better code readability and documentation too.

Most look good to me, I have some minor comments.
- @sauk2 could you have a look at the `unsubscribe` function in the `ProtocolAdapter`?
- It will be great if you can add a test for the teardown, but it seems that it is quite hard to check whether the topic is unsubscribed from the `ProtocolAdapter` and `MqttClient` API?

vda5050_master/include/vda5050_master/agv/agv.hpp
stevelimyh: Thanks for the suggestion.  Agreed, `weak_ptr` is better, eliminates the chance of the silent dangle. Boilerplate isn't bad: master inherits `enable_shared_from_this<VDA5050Master>`, `parent_` becomes `weak_ptr<VDA5050Master>`, and dispatch sites use `parent_.lock()` before use.

vda5050_execution/test/unit/test_protocol_adapter.cpp
```diff
@@ -147,6 +147,15 @@ TYPED_TEST(ProtocolAdapterTest, SubscribeMessage)
...
 
+TYPED_TEST(ProtocolAdapterTest, UnsubscribeMessage)
+{
+  EXPECT_CALL(
+    *this->mock_, unsubscribe(testing::StartsWith(this->topic_prefix_)))
+    .Times(1);
+
+  this->adapter_->template unsubscribe<TypeParam>();
+}
+
```
sauk2, line 200: This current test checks if any error is thrown if unsubscribe is called without an actual subscription. 

I would suggests renaming this test `UnsubscribeWithoutSubscription` (or something like that and adding another test with the name `UnsubscribeMessage`. 

The new test should do the following:
- Start a subscription
- Captured handler should receive a sent message
- Unsubscribe
- Captured handler shouldn't receive any more messages

vda5050_master/include/vda5050_master/agv/agv_impl.hpp
```diff
@@ -0,0 +1,69 @@
...
+void create_subscription(
+  AGV* agv, std::function<void(const MsgType&)> handler, QosLevel qos)
+{
+  agv->protocol_adapter_->template subscribe<MsgType>(
+    [agv, handler = std::move(handler)](
+      MsgType msg, std::optional<vda5050_types::Error> error) {
+      if (error.has_value())
+      {
+        VDA5050_ERROR(
+          "[AGV] Failed to parse message for {}: {}", agv->agv_id_,
```
sauk2: I think there is an unaccounted case here. What happens if the `ProtocolAdapter` outlives the `AGV` instance? Since the `ProtocolAdapter` still holds the callback, the lambda will still run when a message arrives and any attempt to access `agv->agv_id_` will cause undefined behaviour.

Solution would be to either pass in as `std::weak_ptr<AGV>` or since you are only using the `agv_id_`, just capture that as a copy into the lambda.

sauk2 (changes requested):
Thanks for the PR! I have provided my comments

vda5050_master/include/vda5050_master/agv/agv.hpp
Briancbn: I don't think `enable_shared_from_this<VDA5050Master>` is needed here. This is typically needed under the scenario below commonly found in smart pointer builder pattern.

```cpp
class MyClass : public std::enable_shared_from_this<VDA5050Master>
{
public:
  std::shared_ptr<VDA5050Master> func(int) {
    // do something
    ...
    return shared_from_this()  // return a share pointer to itself.
  }
}

std::shared_ptr<MyClass> my_class = 
  std::make_shared<MyClass>()
    .func(1)
    .func2(2);
```

vda5050_master/include/vda5050_master/agv/agv_impl.hpp
Briancbn: I believe this is cleaned up through `unsubscribe` in the destructor?

sauk2: Yes, true but the unsubscription can still fail according to the Paho client code,

```cpp
void PahoMqttClient::unsubscribe(const std::string& topic)
{
  try
  {
    client_->unsubscribe(topic)->wait();
    std::lock_guard<std::mutex> lock(handler_mutex_);
    handlers_.erase(topic);
  }
  catch (const mqtt::exception& e)
  {
    VDA5050_ERROR_STREAM("MQTT unsubscription failed: " << e.get_message());
  }
}
```

In case it does, the `ProtocolAdapter` might still keep getting messages.

P.S. We do need a better way to propagate errors from MQTT client

Briancbn: Thanks I understand the concern here.

vda5050_master/include/vda5050_master/agv/agv.hpp
stevelimyh: Addressed in this [commit](https://github.com/ros-industrial/vda5050_client/pull/45/commits/afb9a593425c71dd35a20f9c3955a48d0de80bf2)

I might be missing something, but I think we need it for `weak_from_this()` (not `shared_from_this()`) — that’s how `master.cpp::onboard_agv()` gets a `weak_ptr<VDA5050Master>` to pass to AGV. Without `enable_shared_from_this`, that method isn’t available.

Is there a different design you have in mind?

vda5050_execution/test/unit/test_protocol_adapter.cpp
stevelimyh: Addressed in this [commit](https://github.com/ros-industrial/vda5050_client/pull/45/commits/8df33224c88e9e536ffe38f2a5f6c842c61667eb).  

Renamed the existing test to `UnsubscribeWithoutSubscription` (applied across all 6 message types), and added a new `UnsubscribeMessage` lifecycle test (applied across the 3 default-constructible types) following your suggested flow:

- Start subscription
- Dispatch message → assert callback fires
- Unsubscribe
- Re-dispatch → assert callback does not fire

To ensure the “shouldn’t receive any more messages” condition holds, `ProtocolAdapter` now maintains a per-T active flag captured by the wrapper. `unsubscribe<T>()` flips this flag to false, so any in-flight or subsequent dispatch becomes a no-op at the adapter layer.

Briancbn (approved):

vda5050_master/include/vda5050_master/agv/agv_impl.hpp
stevelimyh: Addressed in this [commit](https://github.com/ros-industrial/vda5050_client/pull/45/commits/5a5631775c6b30e7fa02cb00378a7fdae4830c5a).

Going with your `weak_ptr<AGV>` suggestion. Lambda captures `weak_from_this()` and `lock()`s before deref, so AGV destruction results in a no-op.

Together with active flag in this [commit](https://github.com/ros-industrial/vda5050_client/pull/45/commits/8df33224c88e9e536ffe38f2a5f6c842c61667eb) which covers the Paho failure case.

sauk2 (approved):

sauk2:
I will merge this in so we can move forward with the package refactor.

@stevelimyh Thanks for all the great work in getting this PR completed and @tanjpg for getting all the initial implementation done!

Issue #41, opened 2026-04-18
Add retain flag to method to set will in the MQTT client


sauk2:
## Summary

The current `set_will` method always uses a hardcoded `retain = true` value internally. This removes flexibility and makes it impossible to control the retain behavior from the caller.

Propose to update the `set_will` method to accept a retain flag as a parameter instead of hardcoding it.

sauk2:
Done as part #48

PR #48, opened 2026-05-15
feat(mqtt)!: changes to mqtt client and protocol adapter 

### LINKED ISSUE #39: Refactor MQTT client ownership in protocol adapter
## Summary

The protocol adapter currently accepts a `std::shared_ptr` to the MQTT client. Since both the protocol adapter and MQTT client are specific to an AGV, sharing the client across multiple adapters can lead to unclear ownership and potential misuse.

The ownership model needs to be refactored to better reflect intended usage. 

Tasks:
- Introducing a factory function that returns `std::unique_ptr` for protocol adapter use. This will be alongside the existing factory function that returns a `std::shared_ptr`.
- Updating the protocol adapter to accept an `std::unique_ptr`
- Updating protocol adapter tests
- Fixing existing usage patterns

sauk2:
Closes #39 #41

## Summary

This PR introduces breaking changes to the MQTT components and updates the `ProtocolAdapter`.

The main change is the addition of helper default for unique MQTT client instances for AGV clients, which aligns with the intended one-will-per `ProtocolAdapter` design.

Changes:
- Introduced a new default factory method to create unique MQTT client instances in the MQTT client and Paho implementation.
- The `ProtocolAdapter` can now use the new method to create and owns its own MQTT client instead of sharing a global instance. This change ensures isolation between adapters of AGV clients, as each MQTT client inside the adapter can only handle a single will.


Briancbn:
> The main change is the removal of shared MQTT client instances in favor of per-adapter clients, which aligns with the intended one-topic-per-ProtocolAdapter design. As a result, the master initialization and shutdown flow has been updated to reflect this new ownership model.

Since the `ProtocolAdapter` is also used in the `master`, changing it to `unique_ptr` will prevent `master` from sharing the MQTT client instance. In the case of `master`, sharing the client instance may be a desired behavior.

Instead of enforcing "unique" for `ProtocolAdapter`, I suggest that the "unique" or "shared" aspect can be enforce at a layer on top of the protocol adapter. In the case of `master` it is "shared", in the case of `client` it is "unique". If we follow this behavior, we should leave the `ProtocolAdapter`, as is for compatibility with both.

sauk2:
> > The main change is the removal of shared MQTT client instances in favor of per-adapter clients, which aligns with the intended one-topic-per-ProtocolAdapter design. As a result, the master initialization and shutdown flow has been updated to reflect this new ownership model.
> 
> Since the `ProtocolAdapter` is also used in the `master`, changing it to `unique_ptr` will prevent `master` from sharing the MQTT client instance. In the case of `master`, sharing the client instance may be a desired behavior.
> 
> Instead of enforcing "unique" for `ProtocolAdapter`, I suggest that the "unique" or "shared" aspect can be enforce at a layer on top of the protocol adapter. In the case of `master` it is "shared", in the case of `client` it is "unique". If we follow this behavior, we should leave the `ProtocolAdapter`, as is for compatibility with both.

I have made this change. Now the actual `ProtocolAdapter` is templatized and the specializations are namespaced as `master` and `client` respectively.

vda5050_core/test/execution/test_protocol_adapter_unique_client.cpp
```diff
@@ -0,0 +1,188 @@
...
+  {
+    interface_ = "uagv";
+    version_ = "v2";
+    manufacturer_ = "ROS-I";
+    serial_number_ = "S001";
+
+    auto mock = std::make_unique<MockMqttClient>();
+    mock_ = mock.get();
+
+    adapter_ = ProtocolAdapter::make(
```
Briancbn: I don't think we need the template for `ProtocolAdapter`. I suggest for it to always takes in the the `MqttClientInterface` as a shared pointer. This can offer the maximum flexibility and support all our use case. If we want the unique implementation, we simply pass in the `unique_ptr<MqttClientInterface>` in and the `ProtocolAdapter` can store it as a `shared_ptr` with only 1 instance.

If you use the macro that I suggested below. We can easily provide `make_unique` and `make_shared` for the `ProtocolAdapter` with minimal boilerplate code. In this way, the test code here should still work in the same way.

vda5050_core/include/vda5050_core/transport/paho_mqtt_client.hpp
```diff
@@ -88,7 +88,16 @@ class PahoMqttClient : public MqttClientInterface
   /// \param client_id ID of the MQTT client
   ///
   /// \return Shared pointer to Paho MQTT client
-  static std::shared_ptr<PahoMqttClient> make(
+  static std::shared_ptr<PahoMqttClient> make_shared(
```
Briancbn, line 91: You can check out the [macro example here](https://github.com/ros-industrial/rmf2_scheduler/blob/main/rmf2_scheduler/include/rmf2_scheduler/macros.hpp), I referenced this from `rclcpp`. It can provide `make_shared` and `make_unique` with minimal boilerplate code.

Briancbn (commented):
I am not really a fan of introducing template here in the constructor. Below is my suggestion.

vda5050_core/test/execution/test_protocol_adapter_unique_client.cpp
sauk2: Reverted most of the constructor changes. Since we are the only ones using this class for now, let's just handle the passing of the MQTT client as share or unique ourselves. Once we start more development on the AGV client, we can think about possibly enforcing the uniqueness of the MQTT client for AG client-side code

vda5050_core/include/vda5050_core/transport/paho_mqtt_client.hpp
sauk2: I was thinking of creating an issue for this so we can work on this later

Briancbn (approved):

PR #50, opened 2026-05-22
feat: add method to suspend engine indefinitely

### LINKED ISSUE #40: Add suspend API without timeout for indefinite suspension
## Summary

The current `suspend_for(timeout, predicate = nullptr)` API in the engine requires a timeout and an optional predicate. This makes it unsuitable for cases where the caller needs to suspend execution indefinitely until a condition is met, without enforcing an artificial timeout.

Propose adding a new API `suspend(predicate)` requiring a predicate. This is to allow the engine to suspend without a timeout and resumes only when the predicate is satisfied.

This avoids workarounds like using very large timeouts and makes the intent clear.

sauk2:
Closes #40 

## Summary

This PR adds a new method to `Engine` to suspend it indefinitely without a timeout. The new `suspend` method takes in only a predicate and the existing `suspend_for` methods still remains unchanged.

### DIFF
--- a/vda5050_core/include/vda5050_core/execution/engine.hpp
+++ b/vda5050_core/include/vda5050_core/execution/engine.hpp
@@ -23,6 +23,7 @@
 #include <functional>
 #include <memory>
 #include <mutex>
+#include <optional>
 #include <unordered_map>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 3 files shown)


Briancbn (approved):
lgtm

PR #53, opened 2026-05-26
feat(client): add concrete execution data types

eileentyz:
## Summary
Adds the foundational core data structures and type-indexing interfaces for the order-based execution layer to vda5050_core. These additions establish the information carriers and identity parameters required for the incoming publish path.

## What's Included
* **OrderUpdate** — Struct wrapping `types::Order` to safely transport incoming fleet orders as an inbound carrier.
* **ConfigResource** — Struct to store and protect persistent AGV identity parameters and configurations.
* **OrderTypesTest** — Implements Google Mock unit tests.


Briancbn:
Hi @eileentyz, thanks for working on this. Although the same data type can potentially be used in `master` as well, I suggest to put these under `client` and `vda5050_core::client` namespace first.

Briancbn (approved):

vda5050_core/include/vda5050_core/client/resources/config.hpp
sauk2, line 1: I would recommend separating the individual structs based on their base types.

You could have individual folders for each base type as follows,

```bash
vda5050_core
    └── include
        └── vda5050_core
            └── client
                 ├── events
                 ├── resources
                 │ └── config.hpp
                 └── updates
                    └── order.hpp       # all order related update structs
```

vda5050_core/include/vda5050_core/client/order_types.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+/// \brief Persistent identity of this AGV on the VDA5050 interface.
+///
+/// Stored once as a `ResourceBase` in the Context. It mirrors the topic
+/// identity used by the `ProtocolAdapter`
+/// (`<interface_name>/<version>/<manufacturer>/<serial_number>/...`) and lets a
+/// Strategy reason about which AGV it is acting for.
+
+// Stores the AGV's identity/config information
+struct ConfigResource
```
sauk2: Minor comment on the naming for this struct. `ConfigResource` sounds broad, could we instead name it `ProtocolConfigResource` or something similar so that there is less confusion when we add more resources containing configurations?

vda5050_core/test/client/test_order_types.cpp
sauk2, line 1: Would be good to move this under `test/client/` instead.

vda5050_core/include/vda5050_core/client/order_types.hpp
```diff
@@ -0,0 +1,90 @@
...
+: public execution::Initialize<ConfigResource, execution::ResourceBase>
+{
+  /// \brief Interface name, e.g. "uagv".
+  std::string interface_name;
+
+  /// \brief Protocol/major version segment, e.g. "v2".
```
sauk2: We should enforce using the full semantic version here instead of the version style used in topics. For now, we can just change documentation as follows,

```suggestion
  /// \brief Protocol/major version segment, e.g. "2.0.0".
```

sauk2: Later we can harden this type as described in this issue https://github.com/ros-industrial/vda5050_core/issues/61

sauk2 (changes requested):
Thanks for the great work! I just have a few minor comments after which we can merge this in!

sauk2 (approved):

sauk2:
@eileentyz Great job! I am merging this in :tada:

PR #54, opened 2026-05-26
feat(layout): add LIF topology graph and JSON loader

stevelimyh:
## Summary

Adds LIF (Layout Interchange Format) layout types, a queryable topology `Graph`, and a JSON loader to `vda5050_core` as a standalone, unit-tested module. LIF is the standard facility-layout exchange format; the `Graph` is the master's internal queryable view that the order validators check against in later PRs. Not yet wired into the publish path.

### `vda5050_core/layout/` — LIF types (standard layout format)

- `Lif` — root document: format metadata + one or more `Layout`s.
- `Layout` — a single facility map: `Node`s, `Edge`s, `Station`s.
- `Node` / `Edge` / `Station` / `Trajectory` — geometry + identity types.
- `LayoutAction`, `LoadRestriction`, `VehicleTypeNodeProperty`, `VehicleTypeEdgeProperty` — per-node/edge attributes (allowed actions, per-vehicle traversal properties, load constraints).

These mirror the LIF spec and are distinct from the Order-wire `types::Node` / `types::Edge`.

### `Graph` — master's queryable topology (derived)

- Built from a `Layout`; provides **O(1)** node / edge / station lookup (hash-backed indices) plus adjacency lists for traversal.
- Lookups by id via `find_*`; layout-level lookup stays O(N_layouts) (N ≤ ~5).

### `layout_loader` — JSON ingestion

- Loads a LIF document from JSON via `load_from_file` / `load_from_json`, returning a `LayoutLoadResult` (`std::variant<Success, Failure>` — a valid LIF *or* errors, never both; no partial load).
- Categorical error reporting:
- File / parse errors
- Missing required fields
- Duplicate IDs
- Dangling edge references
- Empty layout

### `layout_serialization` / `validate_layout`

- `nlohmann_json` (de)serialization for all LIF types (round-trip tested).
- Structural validation enforcing the error categories above.

## Tests

`test/layout/` — `test_unit_graph` and `test_unit_loader` (lookup correctness, round-trip serialization, every load-error category).


Briancbn:
Hi @stevelimyh, I have some suggestion the map format. 

Since your map format is very similar to [VDMA - LIF (Layout Interchange Format)](https://www.vdma.eu/documents/34570/3317035/FuI_Guideline_LIF_GB.pdf)'s `Layout`. I think we should just create the `Layout` data structure as specified or we use the format in its entirety like section 8.3. Let me know your thoughts.
 
I suggest to introduce `Map` or `Layout` as serializable types similar to `Order`, but specific to `master` for now, instead of program it as `parse` functions.

stevelimyh:
> Hi @stevelimyh, I have some suggestion the map format.
> 
> Since your map format is very similar to [VDMA - LIF (Layout Interchange Format)](https://www.vdma.eu/documents/34570/3317035/FuI_Guideline_LIF_GB.pdf)'s `Layout`. I think we should just create the `Layout` data structure as specified or we use the format in its entirety like section 8.3. Let me know your thoughts.
> 
> I suggest to introduce `Map` or `Layout` as serializable types similar to `Order`, but specific to `master` for now, instead of program it as `parse` functions.

Thanks @Briancbn for the LIF doc pointer. Agreed on the serializable type approach.

Let’s go with full LIF section 8.3 alignment instead of a LIF-shaped custom format. I’ll revise this to introduce Layout as serializable master-side types, similar to Order, and update the loader around those types.

sauk2:
One architectural question I have is whether the layout/LIF model should live under `master`.

Based on the document, the topology description appears to be a configuration supplied by the system integrator rather than something owned by the master itself. While the initial consumer may be the master for now, I could also see the same data structures and loader being useful on the client side or in other modules.

Would it make sense to place the layout representation and loader in a more neutral namespace (e.g. `layout`, `map` or `topology`) and have master functionality depend on it instead? This could make the ownership of the data model clearer and allow the same map representation to be reused across both master and client components if needed in the future.

vda5050_core/src/layout/layout.cpp
```diff
@@ -0,0 +1,49 @@
...
+  return it == edges.end() ? nullptr : &(*it);
+}
+
+const Station* Layout::find_station(const std::string& id) const
+{
+  auto it = std::find_if(
+    stations.begin(), stations.end(),
+    [&id](const Station& s) { return s.station_id == id; });
+  return it == stations.end() ? nullptr : &(*it);
+}
```
sauk2: Currently all these functions perform linear searches with complexity O(N). Do we expect these lookups to be used heavily by future logic? If so, it may be worth considering an indexed lookup structure

vda5050_core/include/vda5050_core/master/layout/layout_loader.hpp
```diff
@@ -0,0 +1,76 @@
...
+struct LayoutLoadError
+{
+  LayoutLoadErrorType type;
+  std::string description;
+};
+
+struct LayoutLoadResult
+{
+  std::optional<LIF> lif;
+  std::vector<LayoutLoadError> errors;
```
sauk2: Just a design question - do we expect partial loading in the future or is the intent that loading is always either a valid LIF or a collection of errors? The current implementation appears to enforce the second case, so I was curious whether the optional LIF + errors representation was chosen to leave room for future scenarios

sauk2 (commented):
Thanks for putting effort into introducing a map representation! 

I just have some initial questions about the design before I go through the validation code.

vda5050_core/src/layout/layout.cpp
stevelimyh: Yes, validators in upcoming PRs (graph validation, traversability, order-step lookups) will hit these heavily. Lookups now go through the new vda5050_core::layout::Graph class.

vda5050_core/include/vda5050_core/master/layout/layout_loader.hpp
stevelimyh: The intent is either valid LIF or errors, never both, no partial-load planned. Rewrote LayoutLoadResult around std::variant<Success, Failure> so the two states are mutually exclusive at the type level

vda5050_core/include/vda5050_core/layout/graph.hpp
```diff
@@ -0,0 +1,256 @@
...
+  // node IDs and properties in different layout order are NOT equal.
+  bool operator==(const Graph& rhs) const;
+  bool operator!=(const Graph& rhs) const;
+
+  // Graphviz DOT format. Edge labels are edge IDs.
+  void dump_dot(std::ostream& os) const;
+
+private:
+  Graph() = default;
+  void rebuild_indices_();
```
Briancbn: Follow the convention no underscore at the back. `rebuild_indices`.

vda5050_core/include/vda5050_core/layout/graph.hpp
```diff
@@ -0,0 +1,256 @@
...
+{
+public:
+  using Ptr = std::shared_ptr<Graph>;
+  using ConstPtr = std::shared_ptr<const Graph>;
+
+  // Factory; takes ownership of `lif`. Throws std::invalid_argument if the
+  // input contains duplicate node, edge, station, or layout IDs (same
+  // uniqueness invariant validate_layout enforces). Callers using
+  // load_from_file / load_from_json already get a validated LIF; this guard
+  // protects in-memory or test-constructed inputs.
```
Briancbn: Use Doxygen compatible style?

vda5050_core/include/vda5050_core/layout/layout_loader.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+// Load a single LIF file. Node, edge, station, and layout IDs are required
+// globally unique across the loaded LIF. Multi-vendor reconciliation (where
+// different vendor LIFs may legitimately reuse IDs and must be remapped by
+// the master) is out of scope; loading several vendor LIFs as one will yield
+// DUPLICATE_ID errors by design.
+LayoutLoadResult load_from_file(const std::string& path);
+LayoutLoadResult load_from_json(const nlohmann::json& json);
+
+std::vector<LayoutLoadError> validate_layout(const LIF& lif);
```
Briancbn: Separate into another files following the `validate_layout.cpp`

Briancbn (approved):
Good job! looks really good.

I have some small aesthetic changes, but other than that we are good to go.

Briancbn:
@stevelimyh please follow this style for the API comment
https://github.com/ros-industrial/vda5050_core/blob/a9d8d66a46f0aac95286c4a9b8905442c8c455e9/vda5050_core/include/vda5050_core/logger/logger.hpp#L62-L65

stevelimyh:
> @stevelimyh please follow this style for the API comment
> 
> https://github.com/ros-industrial/vda5050_core/blob/a9d8d66a46f0aac95286c4a9b8905442c8c455e9/vda5050_core/include/vda5050_core/logger/logger.hpp#L62-L65

Thanks @Briancbn. Updated the layout API comments to the `\brief ` style per `logger.hpp`. Could you take another look?

Briancbn:
@sauk2 please help take a look again.

vda5050_core/include/vda5050_core/layout/layout_loader.hpp
```diff
@@ -0,0 +1,87 @@
...
+#include "vda5050_core/layout/lif.hpp"
+#include "vda5050_core/layout/validate_layout.hpp"
+
+namespace vda5050_core::layout {
+
+/// \brief Result of a layout load: exactly one of {LIF, errors} is populated.
+///
+/// Construction is funnelled through success() / failure() factories so an
+/// invalid state cannot be represented.
+class LayoutLoadResult
```
sauk2: `LayoutLoadResult` seems fairly heavy for a type that represents either a successfully loaded LIF or a set of errors. The variant wrapper, factory functions, methods to access the LIF and move APIs add some complexity in usage. Callers must remember to check `ok()` before accessing the stored value. I feel a simpler result structure could provide the same functionality with a minimal set of APIs.

Something like
```cpp
struct LayoutLoadResult
{
  std::optional<LIF> lif;
  std::vector<LayoutLoadError> errors;

  explicit operator bool() const
  {
    return errors.empty();
  }
};
```
This could take the complexity away from the user and reduce the risk of throwing a `std::bad_variant_access`

vda5050_core/src/layout/graph.cpp
```diff
@@ -0,0 +1,593 @@
...
+    }
+  }
+
+  auto g = Ptr(new Graph());
+  g->lif_ = std::move(lif);
+  g->rebuild_indices();
+  return g;
+}
+
+void Graph::rebuild_indices()
```
sauk2, line 46: I have noticed this function is used in other mutation functions like add, delete, update, prune which will make the operations O(N+E+S). If any of those functions are called inside a loop then it would make the code run at probably O(N^2). Do you think it's possible to find an efficient way to perform mutations? 

One way you could do is make sure the graph is not rebuilt at every mutation operation and instead do it after performing all mutations. For example, you can provide an `add_node()` method and `add_node_and_rebuild_indices()` method while also exposing `rebuild_indices()` so users can decide when they want to rebuild the entire graph.

vda5050_core/src/layout/graph.cpp
```diff
@@ -0,0 +1,593 @@
...
+
+#include <map>
+#include <ostream>
+#include <stdexcept>
+#include <unordered_set>
+#include <utility>
+
+namespace vda5050_core::layout {
+
+Graph::Ptr Graph::from_lif(LIF lif)
```
sauk2, line 32: I noticed that you are also check for node/edge/station duplication in `validate_layout()`

vda5050_core/src/layout/graph.cpp
```diff
@@ -0,0 +1,593 @@
...
+}
+
+const Layout* Graph::find_layout(const std::string& layout_id) const noexcept
+{
+  for (const auto& layout : lif_.layouts)
+  {
+    if (layout.layout_id == layout_id) return &layout;
+  }
+  return nullptr;
+}
```
sauk2, line 161: Can you also handle layouts like you do for nodes/edges/stations? Right now it's O(N) here.

vda5050_core/src/layout/graph.cpp
```diff
@@ -0,0 +1,593 @@
...
+{
+  return station_index_.find(id) != station_index_.end();
+}
+
+const Node& Graph::get_node(const std::string& id) const
+{
+  const auto* n = find_node(id);
+  if (n == nullptr)
+  {
+    throw std::out_of_range("Graph::get_node: id '" + id + "' not found");
```
sauk2, line 107: You throw an exception for `get_node()` but for `find_node()` you return a `nullptr`. What do you think about unifying the return type for all functions, either throw exceptions or return a `nullptr`?

vda5050_core/src/layout/graph.cpp
Briancbn: Sidetrack here, I am thinking the `Graph` should be per layout instead of 1 graph for all layouts.

vda5050_core/src/layout/graph.cpp
Briancbn: > One way you could do is make sure the graph is not rebuilt at every mutation operation and instead do it after performing all mutations. For example, you can provide an `add_node()` method and `add_node_and_rebuild_indices()` method while also exposing `rebuild_indices()` so users can decide when they want to rebuild the entire graph.

I suggest not to introduce any additional member functions here, but create some anonymous / private functions to optimize the `add_node` default behavior.

vda5050_core/src/layout/graph.cpp
Briancbn: This looks intentional to me though @stevelimyh. I think this is also quite a common practice in the std library too where `[]` and `at` throws and `find` doesn't

vda5050_core/include/vda5050_core/layout/layout_loader.hpp
stevelimyh: Reverted to the pre-variant struct (`std::optional<LIF>` + `std::vector<LayoutLoadError>`, two fields + `operator bool`) in 3caaa5f. Bool guards `lif.has_value() && errors.empty()` so a successful check always has a usable LIF.

vda5050_core/src/layout/graph.cpp
stevelimyh: Done in 96a2595.

Each mutation now updates the indices in place via anonymous-namespace helpers (`erase_swap`, `rekey`, `vec_erase_value`, `vec_replace_value`), so `rebuild_indices()` is no longer called per mutation and is only used once during `from_lif()` construction.

Resulting per-operation cost (`deg` = node edge count; `K` = station interaction-node count):

- `add_*` → `O(1)`
- `delete_node` → `O(1)`, `delete_edge` → `O(deg)`
- `update_*_id` → `O(deg/refs)`, `*_station` → `O(K)`
- `prune` → `O(N + S)` (scan all nodes + station refs; swap-and-pop deletion is `O(1)`)

vda5050_core/src/layout/graph.cpp
stevelimyh: Thanks for pointing this out. `from_lif` had its own copy of the uniqueness check. Removed it in 1d13aac; `from_lif` now calls `validate_layout()` directly, so duplicate detection lives in one place.

vda5050_core/src/layout/graph.cpp
stevelimyh: Done in 0270116. Added `layout_index_` (`layout_id` → index), similar to the node/edge/station indices, so `find_layout()` is now `O(1)`. This also allowed me to remove the duplicate `find_layout_index()` helper.

On the per-layout graph idea: I can see the appeal, but I'm currently leaning towards keeping this as a single graph. Node/edge/station IDs are unique across all layouts, so we still need global lookup/indexing regardless of how the graph is sliced.

The LIF spec also allows cross-layout edges (`endNodeId` can be located in another layout), so a single graph lets us treat those as ordinary edges. With per-layout graphs, we would likely need an explicit transition table or similar mechanism to handle them.

Since the master validates orders rather than plans routes, the main need here is lookup + local connectivity checks, which the single-graph model serves directly. That said, happy to revisit if you think per-layout graphs would be cleaner here.

vda5050_core/src/layout/graph.cpp
stevelimyh: Yes, this is intentional.

`get_node()` is for when presence is an invariant: if the ID is missing, that indicates a bug/internal inconsistency, so it throws.

`find_node()` is for when absence is a normal outcome to branch on, so it returns `nullptr`.

On the master side:
- `find_node()` is useful during order validation, where a referenced node may legitimately not exist in the loaded map and should become an order rejection.
- `get_node()` is useful after validation has passed, where presence is guaranteed. If it is missing at that point, throwing is the right signal.

sauk2 (approved):
Thanks for the changes! I will let @Briancbn give a last pass

Briancbn:
Lgtm🚀

PR #55, opened 2026-05-26
feat(master): add schema and pre-send validators

stevelimyh:
## Summary

Adds the first group of outgoing-message validators to `vda5050_core`: message-shape validation, AGV-readiness checks.

These validators are implemented as stateless free functions returning `order_utils::ValidationResult`. They are standalone library code, exercised by unit tests, and are not yet wired into the publish path.

## What's Included

- `schema_validator` — structural field checks per message type:
  - Order
  - InstantActions
  - State
  - Connection
  - Factsheet
  - Visualization

- `pre_send_validator` — AGV-readiness gate using a lock-free `PreSendContext` snapshot.
  - Checks connection, operational state, operating mode, paused state, e-stop, and position initialization.

- Supporting additions:
  - `SchemaValidationError`
  - `PreSendValidationError`
  - `SupportedSchemaVersions`


vda5050_core/include/vda5050_core/master/standard_names.hpp
```diff
@@ -62,6 +63,13 @@ constexpr QosLevel InstantActionsQos = QosLevel::AtMostOnce;
 constexpr int ConnectionHeartbeatInterval = 15;  // seconds
 constexpr int StateHeartbeatInterval = 30;       // seconds
 
+/// \brief VDA5050 protocol versions accepted by the schema validator.
+///
+/// The master rejects (outgoing) or drops (incoming) any message whose
+/// header.version is not in this set. Ships with 2.0.0 only; add a 2.1.0
+/// entry when that migration lands.
+inline const std::vector<std::string> SupportedSchemaVersions = {"2.0.0"};
```
sauk2, line 71: For now, we can probably proceed with this, but I was thinking of having a more secure way of handling versions as described here https://github.com/ros-industrial/vda5050_core/issues/61

vda5050_core/include/vda5050_core/master/validation/schema_validator.hpp
```diff
@@ -0,0 +1,58 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef VDA5050_CORE__MASTER__VALIDATION__SCHEMA_VALIDATOR_HPP_
+#define VDA5050_CORE__MASTER__VALIDATION__SCHEMA_VALIDATOR_HPP_
+
+#include "vda5050_core/order_utils/validation_result.hpp"
```
sauk2: I have a question regarding the location of `ValidationResult`. Since it's being used as a return type across all VDA5050 types, should we transfer it under a different common namespace like `vda5050_core::errors`?

vda5050_core/include/vda5050_core/validation/content_validator.hpp
sauk2, line 1: The term "schema validator" feels somewhat misleading here. I would normally expect a schema validator to validate a message against the [VDA5050 schema definition](https://github.com/VDA5050/VDA5050) itself by enforcing required fields, types, enum values, constraints, etc.

The current implementation performs semantic checks on already constructed C++ messages, such as non-empty IDs. Would it make sense to use different terminology instead?

vda5050_core/src/master/validation/schema_validator.cpp
```diff
@@ -0,0 +1,250 @@
...
+        {{::vda5050_core::errors::RefSequenceId,
+          std::to_string(ns.sequence_id)}});
+    }
+  }
+
+  for (const auto& as : state.action_states)
+  {
+    if (as.action_id.empty())
+    {
+      add_error("State.action_states[].action_id must be non-empty", {});
```
sauk2: I think it would be a good idea to follow convention and add an error reference mapped to field here too

sauk2 (changes requested):
Thanks for the PR! I have given some comments regarding the naming and convention followed here

vda5050_core/include/vda5050_core/master/validation/schema_validator.hpp
stevelimyh: Done in 6e44052. Moved ValidationResult to errors/validation_result.hpp and changed its namespace order_utils to errors, since it's the shared return type across all validators (state, connection, factsheet, instant-actions, pre-send), not order-utils-specific.

vda5050_core/include/vda5050_core/validation/content_validator.hpp
stevelimyh: Thanks for pointing this out, I agree "schema" was the wrong word. Renamed in fab6239: `schema_validator` to `content_validator`, `validate_*_schema` to `validate_*_content` (6 functions), and `SchemaValidationError` to `ContentValidationError` ("`contentValidationError`").

vda5050_core/src/master/validation/schema_validator.cpp
stevelimyh: Updated in `6a8b3b9` — the empty `action_id` check now attaches an `actionId` reference (`RefActionId`), matching how the node/edge content checks tag `sequenceId`.

Since `ActionState` does not carry a `sequenceId`, the reference value is the empty `action_id` itself as the best available locator.

Added test coverage for all four references (`orderId`, `orderUpdateId`, `sequenceId`, `actionId`) in `7c00057`.

vda5050_core/src/master/validation/content_validator.cpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+  }
+
+  for (const auto& as : state.action_states)
+  {
+    if (as.action_id.empty())
+    {
+      add_error(
+        "State.action_states[].action_id must be non-empty",
+        {{::vda5050_core::errors::RefActionId, as.action_id}});
```
sauk2: When `action_id` is empty, the attached `RefActionId` is also empty, so it doesn't add much value to the error message. In this case can we attach another identifier like `action_type`?.

vda5050_core/src/master/validation/content_validator.cpp
```diff
@@ -0,0 +1,252 @@
...
+{
+  const auto& supported = SupportedSchemaVersions;
+  if (
+    std::find(supported.begin(), supported.end(), header.version) ==
+    supported.end())
+  {
+    add_error(
+      "header.version '" + header.version +
+        "' is not in SupportedSchemaVersions",
+      {});
```
sauk2:
```suggestion
    add_error(
      "Unsupported header version '" + header.version +
        "'",
      {});
```

vda5050_core/src/validation/content_validator.cpp
sauk2, line 1: I think we can shift both files to a common namespace `vda5050_core::validation`. While you are at it, you can also bring the `order_graph_validator` under the same namespace to have all validation at the same place.

vda5050_core/test/master/test_unit_pre_send_validator.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+TEST(PreSendValidatorTest, MultipleErrorsAccumulate)
+{
+  auto ctx = make_ready_context();
+  ctx.connection_status = vda5050_core::types::ConnectionState::OFFLINE;
+  ctx.last_state->operating_mode = vda5050_core::types::OperatingMode::MANUAL;
+  ctx.last_state->agv_position->position_initialized = false;
+  auto res = validate_pre_send(ctx);
+  EXPECT_FALSE(static_cast<bool>(res));
+  EXPECT_GE(res.errors.size(), 3u);
```
sauk2: I think the size of errors should be deterministic here. We should check for exactly 3 errors, if there is an extra then we should fail the test as the error might be unrelated and added by mistake or a repeated error

```suggestion
  EXPECT_EQ(res.errors.size(), 3u);
```

vda5050_core/src/master/validation/pre_send_validator.cpp
```diff
@@ -0,0 +1,105 @@
...
+  if (ctx.connection_status != vda5050_core::types::ConnectionState::ONLINE)
+  {
+    add_error("AGV connection_status is not ONLINE");
+  }
+
+  if (ctx.operational_state != AGVState::AVAILABLE)
+  {
+    add_error(
+      std::string("AGV operational_state is not AVAILABLE (") +
+      agv_state_name(ctx.operational_state) + ")");
```
sauk2: I would probably recommend using `fmt` here as this project already depends on it

vda5050_core/test/master/test_unit_content_validator.cpp
```diff
@@ -0,0 +1,428 @@
...
+TEST(ContentValidatorTest, MultipleErrorsAccumulate)
+{
+  auto o = make_valid_order();
+  o.order_id = "";               // -> 1 error
+  o.header.version = "9.9.9";    // -> 1 error
+  o.header.manufacturer = "";    // -> 1 error
+  o.nodes.front().node_id = "";  // -> 1 error
+  auto res = validate_order_content(o);
+  EXPECT_FALSE(static_cast<bool>(res));
+  EXPECT_GE(res.errors.size(), 4u);
```
sauk2:
```suggestion
  EXPECT_EQ(res.errors.size(), 4u);
```

sauk2 (changes requested):
Thanks for the changes. This PR is very close to getting merged! Just a few final comments before I approve.

vda5050_core/src/master/validation/content_validator.cpp
stevelimyh: Thanks for another round of review!
Done, addressed in 7a55d8e.

vda5050_core/test/master/test_unit_content_validator.cpp
stevelimyh: Done. Addressed in 8a40f23.

vda5050_core/test/master/test_unit_pre_send_validator.cpp
stevelimyh: Done. Addressed in 8a40f23.

vda5050_core/src/validation/content_validator.cpp
stevelimyh: Done. Addressed in ea0e146.

vda5050_core/src/master/validation/pre_send_validator.cpp
stevelimyh: Done. Addressed in 7a55d8e.

vda5050_core/src/master/validation/content_validator.cpp
stevelimyh: Done. Addressed in 7a55d8e.

sauk2 (approved):
Thanks for the great work! :rocket:

PR #56, opened 2026-05-26
feat(master): add traversability, action-conflict, factsheet, and mode validators

stevelimyh:
> **Stacked PR — depends on #55.** Please review / merge #55 first.
>
> These validators consume `PreSendContext` from #55.

## Summary

Completes the outgoing-validation chain with order / instant-action feasibility checks and an advisory factsheet-vs-layout alignment check.

These validators check whether an AGV can physically execute a command, whether an instant action conflicts with current AGV activity or operating mode, and whether the loaded layout is aligned with the AGV's cached factsheet. Implemented as stateless free functions — standalone library code, not yet wired into the publish path.

## What's Included

* **`traversability_validator`** — validates order feasibility against the loaded `layout::Graph`, AGV position, cached factsheet action capabilities, and protocol array-size limits.

  * Layout/factsheet checks are skipped with a warning when no cached data is available.
  * Node coordinates are not compared; layout agreement is asserted via `map_id`.

* **`action_conflict_validator`** — checks `blocking_type` conflicts against running actions and driving state.

  * Uses distinct error codes: `actionBlockedByDrivingError`, `hardActionBlockedError`.

* **`instant_action_mode_validator`** — applies the operating-mode gate using the predefined instant-scope action allowlist.

  * Rejections are tagged as `modeValidationError`.

* **`factsheet_alignment`** — advisory cross-check of layout speed limits against the AGV factsheet.

  * checks per-edge, per-vehicle-type `max_speed` against `physical_parameters.speed_max` / `speed_min`.


vda5050_core/src/validation/traversability_validator.cpp
```diff
@@ -0,0 +1,397 @@
...
+          "). Layout edges are directed; reverse traversal needs a separate "
+          "edge.",
+        {{::vda5050_core::errors::RefEdgeId, edge.edge_id}});
+    }
+  }
+}
+
+}  // namespace
+
+ValidationResult validate_traversability(
```
sauk2: This function seems to perform several checks - reachability, layout, action capability validation, parameter validation, protocol limit checks. The term "traversbility" doesn't fully describe the responsibility. Maybe would be best to evaluate if we need to separate concerns or rename the validate to better fits its current role

vda5050_core/src/master/validation/instant_action_mode_validator.cpp
```diff
@@ -0,0 +1,83 @@
...
+  return exempt_action_types().count(action_type) != 0;
+}
+
+vda5050_core::order_utils::ValidationResult validate_instant_action_mode(
+  const PreSendContext& ctx, const vda5050_core::types::InstantActions& actions)
+{
+  vda5050_core::order_utils::ValidationResult res;
+
+  // No reported state => unknown mode; pass through.
+  if (!ctx.last_state.has_value()) return res;
```
sauk2: Several validators seem to pass validation entirely when no reported state is present. I am not sure if it's the right move from a safety perspective as this effectively treats unknown state as success instead of indeterminate.

vda5050_core/src/master/validation/traversability_validator.cpp
```diff
@@ -0,0 +1,397 @@
...
+  if (ctx.loaded_graph != nullptr)
+  {
+    validate_graph_integrity(ctx, order, add_error);
+  }
+  else
+  {
+    VDA5050_WARN(
+      "[traversability] No layout loaded; skipping graph-integrity checks for "
+      "order '{}'.",
+      order.order_id);
```
sauk2: This validator currently emits warnings which could repeatedly spam the console with logs. 
There are two error levels `FATAL` and `WARING` so you can change `ValidationResult` to log fatal errors and warnings separately. Then the users can decide if they want to log the warnings or not

vda5050_core/src/master/validation/factsheet_alignment.cpp
```diff
@@ -0,0 +1,87 @@
...
+bool FactsheetAlignmentResult::has_error() const
+{
+  return std::any_of(findings.begin(), findings.end(), [](const auto& f) {
+    return f.severity == AlignmentSeverity::ERROR;
+  });
+}
+
+FactsheetAlignmentResult check_factsheet_alignment(
+  const vda5050_core::layout::Graph& graph,
+  const vda5050_core::types::Factsheet& factsheet)
```
sauk2: All validators seem to return a `ValidationResult` except this. Maybe this can be shifted out of validators or the return type changed

sauk2 (changes requested):
Great work on the PR! Similar to a comment left on #55, I think these validators maybe useful even outside the master so might be a good idea to find a common namespace for them

vda5050_core/src/validation/traversability_validator.cpp
stevelimyh: Agreed, split this into three concern-specific validators in 3fb4ecd.

* `traversability` — reachability to the first node and graph-integrity checks against the loaded layout
* `capability` — action type/scope/blocking and parameter checks against the factsheet
* `protocol_limits` — factsheet array-size limits

Each validator now has its own error code: `traversabilityValidationError`, `capabilityValidationError`, and `protocolLimitError`. The test files were split to match.

The InstantActions path now runs `capability` + `protocol_limits`. `traversability` is order-graph-specific, so that overload was dropped.

vda5050_core/src/master/validation/instant_action_mode_validator.cpp
stevelimyh: Fixed the unsafe pass-through and documented the remaining best-effort case in 2dbb77c:

* `instant_action_mode_validator` no longer passes blindly when no state is available. Unknown mode is now conservative: only exempt recovery/diagnostic actions pass; non-exempt actions are rejected because the validator cannot confirm `AUTOMATIC` / `SEMIAUTOMATIC`.
* `action_conflict_validator` keeps the pass-through, but the comment now states the rationale: without state there is no running-action / driving input to check, and rejecting here would also block recovery actions for a silent AGV.

The non-exempt case is still backstopped by the mode gate above.

vda5050_core/src/master/validation/traversability_validator.cpp
stevelimyh: Done in 91f8ab7. `ValidationResult` now distinguishes fatal errors from warnings via `has_fatal()` / `warnings()`, and validation gates on `has_fatal()` so WARNING advisories do not fail the result.

The previous `VDA5050_WARN` logs for skipped checks are now WARNING-level entries in the result, so the consumer can decide whether to log them.

vda5050_core/src/master/validation/factsheet_alignment.cpp
stevelimyh: Changed in c522eff — `check_factsheet_alignment()` now returns `ValidationResult` like the other validators.

The bespoke `FactsheetAlignmentResult` / `AlignmentFinding` / `AlignmentSeverity` types were removed. Alignment findings now map to WARNING-level `Error` entries, with the alignment codes promoted to `error_codes.hpp`.

This keeps the alignment advisories forwardable through the same result path as the rest of the validator output.

vda5050_core/src/validation/action_conflict_validator.cpp
```diff
@@ -0,0 +1,131 @@
...
+  for (const auto& as : state.action_states)
+  {
+    if (is_active_status(as.action_status))
+    {
+      any_active = true;
+      break;
+    }
+  }
+
+  for (const auto& action : actions.actions)
```
sauk2, line 72: All other validators seem to accumulate errors but this one seems to fail-fast. Would be better to accumulate errors here too

vda5050_core/src/master/validation/instant_action_mode_validator.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+    res.errors.push_back(vda5050_core::errors::create_error(
+      vda5050_core::errors::ModeValidationError,
+      "action_type '" + action.action_type +
+        "' is not on the instant-scope allowlist and the AGV is not "
+        "confirmed to be in AUTOMATIC / SEMIAUTOMATIC operating_mode (master "
+        "must not send driving orders or non-recovery actions in MANUAL / "
+        "SERVICE / TEACHIN, or when the AGV's mode is unknown)",
+      {{vda5050_core::errors::RefActionId, action.action_id}}));
+    return res;  // short-circuit on first failure
```
sauk2: Is there a need to fail-fast? Or can we also let the errors accumulate here?

vda5050_core/include/vda5050_core/order_utils/validation_result.hpp
```diff
@@ -29,13 +31,37 @@ namespace order_utils {
 
 struct ValidationResult
 {
-  /// \brief A vector of error(s) that resulted in an invalid order. Empty if
-  /// order is valid.
+  /// \brief Entries produced by validation. FATAL entries are hard failures
+  /// (see has_fatal()); WARNING entries are advisory. Empty when validation
+  /// passed with nothing to report.
   std::vector<vda5050_core::types::Error> errors;
```
sauk2: I think it would be a good idea to have two vectors here - `fatal_errors` and `warnings`. this would also greatly simplify the `has_fatal`, `has_warnings` functions and `bool` operator

vda5050_core/include/vda5050_core/order_utils/validation_result.hpp
```diff
@@ -29,13 +31,37 @@ namespace order_utils {
...
+  }
+
+  /// \brief Allows use in boolean contexts.
   ///
-  /// \return True if the order is valid, false otherwise.
+  /// \return True iff there are no entries at all (fully clean). Consumers that
+  /// tolerate advisory WARNING entries should gate on has_fatal() instead.
   explicit operator bool() const
   {
     return errors.empty();
```
sauk2: I think this can only check the `fatal_errors` vector. This way the validation will return true if there are only warnings

vda5050_core/include/vda5050_core/order_utils/validation_result.hpp
```diff
@@ -29,13 +31,37 @@ namespace order_utils {
 
 struct ValidationResult
```
sauk2: I would recommend turning `ValidationResult` into a class and making the error vectors private. Then you can have an `add_error` method right here to check the `error_level` and append it to the correct vector.

sauk2 (changes requested):
Thanks for the changes. This PR is almost ready to be merged!

Similar to a comment left on #55, all the validators can be place under a common `vda5050_core::validation` namespace.

vda5050_core/src/validation/action_conflict_validator.cpp
stevelimyh: Thanks for catching this! Addressed in 077a4eb.

vda5050_core/src/master/validation/instant_action_mode_validator.cpp
stevelimyh: Right, no need to fail-fast here. Accumulated instead, addressed in 077a4eb.

vda5050_core/include/vda5050_core/order_utils/validation_result.hpp
stevelimyh: Done in 077a4eb

vda5050_core/include/vda5050_core/order_utils/validation_result.hpp
stevelimyh: Done in 077a4eb

vda5050_core/include/vda5050_core/order_utils/validation_result.hpp
stevelimyh: Done in 077a4eb

sauk2 (approved):
Thanks for the changes! We are ready to merge this :rocket:

vda5050_core/include/vda5050_core/validation/action_conflict_validator.hpp
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+#ifndef VDA5050_CORE__VALIDATION__ACTION_CONFLICT_VALIDATOR_HPP_
+#define VDA5050_CORE__VALIDATION__ACTION_CONFLICT_VALIDATOR_HPP_
+
+#include "vda5050_core/errors/validation_result.hpp"
+#include "vda5050_core/types/instant_actions.hpp"
+#include "vda5050_core/validation/pre_send_validator.hpp"
+
+namespace vda5050_core::validation {
```
Briancbn: I suggest to keep it as the following for header files for better compatibility with C++14.

It is ok to keep it like these in the `.cpp` files.

Please amend those in the previous PR too #55

```cpp
namespace vda5050_core
{
namespace validation
{

}  // namespace validation
}  // namespace vda5050_core
```

vda5050_core/include/vda5050_core/validation/factsheet_alignment.hpp
```diff
@@ -0,0 +1,57 @@
...
+// Speed rule: "speedExceedsCapability" / "speedBelowMinimum" — for every edge's
+// per-vehicle-type max_speed, compare against the factsheet's reported physical
+// speed envelope (physical_parameters.speed_max / speed_min). The v2.0.0
+// factsheet carries no vehicle_type_id, so the check cannot resolve which
+// vehicle-type lane applies to this AGV; it reports every lane and names the
+// vehicle_type_id in the finding, leaving relevance to the integrator. Findings
+// are advisory (WARNING level) and never block onboarding.
+//
+// Deferred (no v2.0.0 factsheet field to check against): footprint-fits-edge,
+// localization tolerance, mapId / mapVersion match.
```
Briancbn: Please avoid comments with a wall of text, the general rule should be 
- document with source code (naming of variables)
- keep the comments directed and succinct. explain along with the lines of code too.

vda5050_core/src/validation/action_conflict_validator.cpp
```diff
@@ -0,0 +1,135 @@
...
+namespace {
+
+using ::vda5050_core::errors::ActionBlockedByDrivingError;
+using ::vda5050_core::errors::create_error;
+using ::vda5050_core::errors::HardActionBlockedError;
+using ::vda5050_core::errors::RefActionId;
+using ::vda5050_core::errors::ValidationResult;
+using ::vda5050_core::types::ActionStatus;
+using ::vda5050_core::types::BlockingType;
+using ::vda5050_core::types::ErrorReference;
```
Briancbn: Please avoid these in the implementations. It is ok to do this in tests. I would recommend to use `errors::create_error` instead in the code.

vda5050_core/src/validation/action_conflict_validator.cpp
```diff
@@ -0,0 +1,135 @@
...
+  // Best-effort pre-screen: this check is state-driven (it reads `driving` and
+  // the running `action_states`). With no reported state there is no input to
+  // screen against, so the conflict cannot be checked here. This is a known
+  // limitation, not a guarantee — the AGV is the runtime authority and enforces
+  // non-conflict itself (every received action gets an actionState; parallel
+  // execution is governed by blockingType). Rejecting on no-state instead would
+  // block recovery actions (cancelOrder, stateRequest) on a silent AGV, which
+  // is worse. This complements the instant-action mode gate, which already
+  // rejects non-exempt actions when the mode is unknown — so this pass-through
+  // cannot admit a non-recovery action on its own.
```
Briancbn: Same as before, please avoid these comments

Briancbn (changes requested):
I have some additional comments.

vda5050_core/include/vda5050_core/validation/action_conflict_validator.hpp
stevelimyh: Thanks for the review! Validation headers (#55 + this PR) converted to nested in 550a4cb .

Similar issues in `layout/` handled in a separate PR #78.

vda5050_core/src/validation/action_conflict_validator.cpp
stevelimyh: Done in 06dda48.

vda5050_core/src/validation/action_conflict_validator.cpp
stevelimyh: Done in 30a6a24.

vda5050_core/include/vda5050_core/validation/factsheet_alignment.hpp
stevelimyh: Done in 30a6a24.

Briancbn: Thanks for the effort. I saw you did an update of the documentation too. Good work!

Briancbn (approved):

PR #58, opened 2026-05-26
feat: master-side event detectors and heartbeat hardening

stevelimyh:
## Summary

Adds master-side event detectors that turn raw AGV `State` and `Connection` messages into discrete transitions for fleet logic, and hardens the existing `HeartbeatListener`.

The detectors are stateless free functions over previous / current snapshots. They are standalone library code, exercised by unit tests, and are not yet wired into `agv` / `master`.

### State Event Detector (`master/state/state_event_detector`)

- Pure previous / current `State` diff helpers for reached-node, error, new-base, mode, paused, driving, and loads transitions.
- `newly_reached_node` is based on `lastNodeId` / `lastNodeSequenceId` advancement; horizon-release flag changes are treated as order updates, not traversal.
- Error transitions are tracked by type, references, and description, so same-type errors from different sources remain distinct.

### Connection Event Detector (`master/connection/connection_event_detector`)

- `detect_connection_transition` edge-detects `ONLINE` / `OFFLINE` / `CONNECTIONBROKEN`, emitting only on change or first message.

### Heartbeat Hardening (`master/heartbeat`)

- Uses `steady_clock` for liveness timing.
- Unifies condition-variable wait / predicate / notify on one mutex for prompt stop handling.
- Hardens teardown ordering and documents the subclassing contract.


Briancbn:
Thanks @stevelimyh for the PR.

> The detectors are stateless free functions over previous / current snapshots. They are standalone library code, exercised by unit tests, and are not yet wired into agv / master.

Is it possible to make use of templates or [the execution library](https://github.com/ros-industrial/vda5050_core/blob/main/vda5050_core/docs/usage.md#21-defining-custom-data) to combine the functions into one?


> Uses `steady_clock` for liveness timing.

I am ok with this change.

There also seem to be a lot of unnecessary changes in this PR, could you revert those?

sauk2:
> Is it possible to make use of templates or [the execution library](https://github.com/ros-industrial/vda5050_core/blob/main/vda5050_core/docs/usage.md#21-defining-custom-data) to combine the functions into one?

I would recommend using `UpdateBase` and then pushing them through an instance of `Provider` to invoke all listener callbacks.

stevelimyh:
> > Is it possible to make use of templates or [the execution library](https://github.com/ros-industrial/vda5050_core/blob/main/vda5050_core/docs/usage.md#21-defining-custom-data) to combine the functions into one?
> 
> I would recommend using `UpdateBase` and then pushing them through an instance of `Provider` to invoke all listener callbacks.

@sauk2 Thanks for pointing this out. Done in a2b494b, the detector now pushes `UpdateBase` transition updates through a shared `Provider`, with listeners registering via `provider->on<...>()`.

Could you take another look when you have time?

vda5050_core/include/vda5050_core/master/state/state_update_detector.hpp
Briancbn, line 1: This is still quite manual to me, but I also can't seem to think of a better way to do this. I am ok to save the better pattern for future consideration.
@sauk2 thoughts

vda5050_core/include/vda5050_core/master/updates/agv_updates.hpp
sauk2, line 1: I think we should align the naming here to avoid confusion. We have a clear distinction in `execution::Base` between,
- `EventBase` for synchronous events (e.g `NavigateToNodeEvent`)
- `UpdateBase` for async changes
- `ResourceBase` for persistent state

The types introduced in this PR are derived from `UpdateBase` and represent the asynchronous updates received from the AGV. So, we should avoid referring to these as “events” in naming or documentation to prevent confusion with `EventBase` and instead call them "updates".

vda5050_core/include/vda5050_core/master/event_detector.hpp
```diff
@@ -0,0 +1,184 @@
...
+///
+/// Each on_state / on_connection call diffs against the previously seen message
+/// and pushes one update per transition through the shared provider, tagged
+/// with this AGV's id. The first message of each kind only seeds the baseline
+/// (a Connection ONLINE is itself the CONNECTED transition).
+///
+/// One AGV's messages must be delivered in arrival order: concurrent calls are
+/// race-free (the snapshot is mutex-guarded) but transition order across them
+/// is not guaranteed.
+class EventDetector
```
sauk2: I think we can remove this class completely and instead replace it with an implementation of `ContextInterface`. The context will already own a `Provider` 

Since the context already provides access to the `Provider`, it would effectively become the single place where incoming state/connection messages can be transformed into updates. You will also be able to store in a snapshot of state for comparison.

sauk2 (changes requested):
Thanks for the PR! I have provided some comments

vda5050_core/include/vda5050_core/master/state/state_update_detector.hpp
sauk2: I agree, but as part of this PR it can be simplified. Since we are already introducing a new class, might as well make it a context that owns a provider

vda5050_core/include/vda5050_core/master/updates/agv_updates.hpp
stevelimyh: Thanks for the review. 

Done, renamed the `UpdateBase`-derived types to use update terminology instead of event terminology: `namespace event` to `namespace update`, `ConnectionEventKind` to `ConnectionTransition`, with files/tests/targets updated to match.

vda5050_core/include/vda5050_core/master/event_detector.hpp
stevelimyh: Done in dc8b4ac. Replaced `EventDetector` with `AGVUpdateContext : execution::ContextInterface`.

`AGVUpdateContext` now owns the `Provider`, centralizes `State` / `Connection` update transformation, stores the previous snapshot for comparison, and caches the latest message per type for `get_update<T>()`.

vda5050_core/include/vda5050_core/master/contexts/agv_update_context.hpp
```diff
@@ -0,0 +1,200 @@
...
+  {
+  }
+};
+
+/// \brief Per-AGV context: turns inbound State/Connection into typed updates
+/// on its Provider, caching the latest of each type for get_update<T>().
+///
+/// on_state / on_connection must be called from one thread; get_update<T>()
+/// is safe from any thread.
+class AGVUpdateContext : public execution::ContextInterface
```
sauk2: Currently there will be one context per AGV, with each context maintaining its own previous state, previous connection, update cache and provider. Instead, I would suggest having a single master-level context that owns all AGV states internally (probably also rename it to something like `MasterContext`).
I think having a single context would avoid creating many duplicated providers and execution contexts as the number of AGVs grows.

The main limitation here would be the current `Provider` API, since all listeners would receive updates from all AGVs and need to filter using `agv_id`, which I think works for now. 

I was thinking of bring in routing support in Provider a separate PR and have create an issue for it - https://github.com/ros-industrial/vda5050_core/issues/79

vda5050_core/include/vda5050_core/master/updates/agv_updates.hpp
sauk2, line 1: Could you separate the updates under an update folder and context stays under the current folder

vda5050_core/src/master/connection/connection_update_detector.cpp
sauk2, line 1: The functions in `connection_update_detector` and `state_update_detector` appear to only support the context's `on_connection() and `on_state()` processing. Since they are stateless helpers and not currently reused elsewhere, it would simplify the design to keep them private to the context?

vda5050_core/src/master/state/state_update_detector.cpp
```diff
@@ -0,0 +1,123 @@
...
+}  // namespace
+
+std::optional<ReachedNode> newly_reached_node(
+  const vda5050_core::types::State& prev,
+  const vda5050_core::types::State& curr)
+{
+  if (curr.last_node_id.empty()) return std::nullopt;
+  if (
+    curr.last_node_id == prev.last_node_id &&
+    curr.last_node_sequence_id == prev.last_node_sequence_id)
```
sauk2: It will trigger a `true` flag on the change in the node related values. This could also happen if state messages arrive out of order with an old message coming later. Then this function will still consider a new node has been reached.

vda5050_core/src/master/state/state_update_detector.cpp
```diff
@@ -0,0 +1,123 @@
...
+namespace update {
+
+namespace {
+
+bool same_error(
+  const vda5050_core::types::Error& a, const vda5050_core::types::Error& b)
+{
+  return a.error_type == b.error_type &&
+         a.error_references == b.error_references &&
+         a.error_description == b.error_description;
```
sauk2: I am not sure if it is a good idea to match error description. Since it could be continuously updated but the actual `errorType` and `errorReferences` would still be the same

sauk2 (changes requested):

vda5050_core/include/vda5050_core/master/contexts/agv_update_context.hpp
stevelimyh: Thanks for the review.

Done in 90c674b. Consolidated the per-AGV `AGVUpdateContext` into one fleet-wide `MasterContext` — it keeps each AGV's baseline in maps keyed by `agv_id` and publishes on the single shared `Provider,` every update tagged with `agv_id`.

vda5050_core/include/vda5050_core/master/updates/agv_updates.hpp
stevelimyh: Done in 893029f

vda5050_core/src/master/connection/connection_update_detector.cpp
stevelimyh: Done in 5e6177a. The detector functions are now in an anonymous namespace inside `master_context.cpp`

vda5050_core/src/master/state/state_update_detector.cpp
stevelimyh: Good catch, fixed in c68e6e1. `on_state` now discards any State older than the last one processed for that AGV before it's compared against the previous State, so a reordered or stale message can't re-fire an event like node-reached. The baseline resets on reconnect so a restarted counter isn't locked out.

vda5050_core/src/master/state/state_update_detector.cpp
stevelimyh: Done in 27469e7. `same_error` now keys identity on `error_type` + `error_references` only. Added tests that a description-only change isn't reported as resolved+reappeared, and that a references change is a new error.

vda5050_core/include/vda5050_core/master/updates/agv_updates.hpp
```diff
@@ -0,0 +1,151 @@
...
+namespace vda5050_core {
+
+namespace master {
+
+/// \brief A node the AGV reports as reached (lastNodeId + lastNodeSequenceId).
+struct ReachedNode
+{
+  std::string node_id;
+  uint32_t sequence_id;
+};
```
sauk2: What do you think about putting the individual fields into `NodeReachedUpdate` instead of having them under a separate struct?

sauk2 (approved):
Just a small comment. Other than that, I am okay with this! Thanks for the great work!

Briancbn (approved):

PR #59, opened 2026-05-28
feat(master): change interface_name to be configurable in vda5050_core::master

Briancbn:
Currently the `InterfaceName` is hard set to [`rmf2` in master](https://github.com/ros-industrial/vda5050_core/blob/develop/vda5050_core/include/vda5050_core/master/standard_names.hpp#L40). However, this should either be `uagv`, which is the only example interface name [mentioned in the VDA5050 docs](https://github.com/VDA5050/VDA5050/blob/release/2.0.0/VDA5050_EN_V1.md#-63-mqtt-topic-levels) or configurable from the master.

This Pull Request implements the latter.

### DIFF
--- a/vda5050_core/src/master/master.cpp
+++ b/vda5050_core/src/master/master.cpp
@@ -87,10 +87,11 @@ bool VDA5050Master::is_connected() const
 
 //=============================================================================
 void VDA5050Master::onboard_agv(
+  const std::string& interface_name,
   const std::string& manufacturer, const std::string& serial_number,
   size_t max_queue_size, bool drop_oldest)
 {
-  std::string agv_id = manufacturer + "/" + serial_number;
+  std::string agv_id = interface_name + "/" + manufacturer + "/" + serial_number;
 
   std::lock_guard<std::mutex> lock(agv_mutex_);
 
@@ -106,8 +107,8 @@ void VDA5050Master::onboard_agv(
   // detecting master destruction cleanly via lock().
   auto agv = std::make_shared<AGV>(
     ProtocolAdapter::make(
-      mqtt_client_, InterfaceName, Version, manufacturer, serial_number),
-    manufacturer, serial_number, max_queue_size, drop_oldest,
+      mqtt_client_, interface_name, Version, manufacturer, serial_number),
+    interface_name, manufacturer, serial_number, max_queue_size, drop_oldest,
     StateHeartbeatInterval, weak_from_this());
 
   // Subscriptions are wired here (not in AGV's ctor) so that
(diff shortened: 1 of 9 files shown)


Briancbn:
Reverted the changes to `agv_id` to prevent the following scenario
- robot with indentical `Manufacturer/SerialNumber`, but different `InterfaceName` exists at the same time.

I also made customization of the `interface_name` opt-in to retain the API compatibility with old.

vda5050_core/src/master/master.cpp
```diff
@@ -120,6 +120,13 @@ void VDA5050Master::onboard_agv(
   VDA5050_INFO("[VDA5050Master] Onboarded AGV: {}", agv_id);
 }
 
+void VDA5050Master::onboard_agv(
+  const std::string& manufacturer, const std::string& serial_number,
+  size_t max_queue_size, bool drop_oldest)
+{
+  onboard_agv("uagv", manufacturer, serial_number, max_queue_size, drop_oldest);
```
sauk2: Instead of hardcoding the interface name here, what do you think about adding in the following in `standard_names.hpp` and using it here?

```cpp
const std::string DefaultInterfaceName = "uagv"; 
```

sauk2 (changes requested):
Thanks for this change! I have provided a minor comment

vda5050_core/test/master/test_integration_master_teardown.cpp
```diff
@@ -89,6 +89,55 @@ TEST_F(MasterTeardownTest, OffboardAgvUnsubscribesAllPerAgvTopics)
...
+      unsubscribe(::testing::HasSubstr("amr/v2/acme/agv-001/factsheet")))
+      .Times(1);
+    EXPECT_CALL(
+      *mock_,
+      unsubscribe(::testing::HasSubstr("amr/v2/acme/agv-001/visualization")))
+      .Times(1);
+
+    master->offboard_agv("acme", "agv-001");
+  }
+}
```
stevelimyh, line 140:
```suggestion
}

```

vda5050_core/test/master/test_integration_master_mqtt.cpp
```diff
@@ -106,6 +108,28 @@ TEST_F(MasterMqttTestFixture, OnboardAGVWhileConnected)
...
   auto agv = master->get_agv(manufacturer_, serial_number_);
   ASSERT_NE(agv, nullptr);
+  EXPECT_EQ(agv->get_interface_name(), interface_name_);
+  EXPECT_EQ(agv->get_manufacturer(), manufacturer_);
+  EXPECT_EQ(agv->get_serial_number(), serial_number_);
+
+  master->disconnect();
+}
+
+TEST_F(MasterMqttTestFixture, OnboardAGVWCustomInterfaceNameWhileConnected)
```
stevelimyh:
```suggestion
TEST_F(MasterMqttTestFixture, OnboardAGVWithCustomInterfaceNameWhileConnected)
```

stevelimyh (commented):
Thanks for the PR, looks good to me overall. Other than @sauk2's request, I only left a few minor nits.

sauk2 (approved):

PR #60, opened 2026-05-29
feat(mqtt): bound MQTT operations with a timeout and harden pub/sub

GameTL:
## Summary

Bounds all blocking MQTT operations with a configurable timeout so a slow or unreachable broker can no longer hang the process, and hardens `PahoMqttClient`'s subscribe/unsubscribe and handler dispatch. Adds an example client and an integration test covering the disconnect path.

The changes are confined to the transport layer (`PahoMqttClient`); the public `MqttClientInterface` is unchanged, so existing `agv` / `master` callers are unaffected.

## Why?
Before client using this mqtt wouldn't be able to `KeyboardInterupt` and need to SIGKILL. This is cause problem when not handle properly.

### Operation Timeout (`transport/paho_mqtt_client`)

- Replaces unbounded `token->wait()` calls with `wait_for_token()`, which blocks for at most `operation_timeout_` (default 3s) across connect, disconnect, publish, subscribe, and unsubscribe.
- Adds `set_operation_timeout()` to configure the bound at runtime.
- `disconnect()` returns promptly: it short-circuits when not connected and passes the timeout through to Paho's `disconnect_options`.

### Subscribe / Unsubscribe & Dispatch Hardening (`transport/paho_mqtt_client`)

- `message_arrived` copies the handler out and releases `handler_mutex_` before invoking it, so a handler that calls back into `subscribe()` / `unsubscribe()` cannot self-deadlock.
- `subscribe()` registers the handler *before* the broker SUBACK and rolls back on failure, so messages arriving between SUBACK and registration are not dropped.
- `unsubscribe()` drops the local handler only once the broker confirms, keeping local state consistent with the broker.
- `disconnect()` no longer mutates `conn_options_`, so reconnect options reused by the next `connect()` are not corrupted.

### Example Client (`examples/mqtt_stay_connected`)

- Connects, subscribes, and holds the connection open until `SIGINT` / `SIGTERM`, then disconnects cleanly — for manually verifying connect/disconnect behaviour against a live broker.

### Tests (`test/transport/test_integration_paho_mqtt_client`)

- Asserts connect/disconnect to an unreachable broker completes within the configured timeout rather than hanging.

### DIFF
--- a/vda5050_core/src/transport/paho_mqtt_client.cpp
+++ b/vda5050_core/src/transport/paho_mqtt_client.cpp
@@ -127,7 +127,8 @@ void PahoMqttClient::set_operation_timeout(std::chrono::milliseconds timeout)
 }
 
 //=============================================================================
-bool PahoMqttClient::wait_for_token(mqtt::token_ptr token, const char* operation)
+bool PahoMqttClient::wait_for_token(
+  mqtt::token_ptr token, const char* operation)
 {
   if (!token) return true;
 
@@ -171,8 +172,7 @@ void PahoMqttClient::disconnect()
     const auto timeout_ms = static_cast<int>(operation_timeout_.count());
     mqtt::disconnect_options options(timeout_ms);
 
-    if (
-      wait_for_token(client_->disconnect(options), "disconnect"))
+    if (wait_for_token(client_->disconnect(options), "disconnect"))
     {
       VDA5050_INFO_STREAM(
         "MQTT client disconnected: " << client_->get_client_id());
--- a/vda5050_core/src/transport/paho_mqtt_client.cpp
+++ b/vda5050_core/src/transport/paho_mqtt_client.cpp
@@ -131,12 +131,11 @@ bool PahoMqttClient::wait_for_token(mqtt::token_ptr token, const char* operation
 {
   if (!token) return true;
 
-  const auto timeout_ms = static_cast<long>(operation_timeout_.count());
-  if (token->wait_for(timeout_ms)) return true;
+  if (token->wait_for(operation_timeout_)) return true;
 
   VDA5050_WARN(
-    "MQTT {} timed out after {} ms for client [{}]", operation, timeout_ms,
-    client_->get_client_id());
+    "MQTT {} timed out after {} ms for client [{}]", operation,
+    operation_timeout_.count(), client_->get_client_id());
   return false;
 }
 
--- a/vda5050_core/src/transport/paho_mqtt_client.cpp
+++ b/vda5050_core/src/transport/paho_mqtt_client.cpp
@@ -80,12 +80,17 @@ void MqttCallback::connection_lost(const std::string& /*cause*/)
 //=============================================================================
 void MqttCallback::message_arrived(mqtt::const_message_ptr msg)
 {
-  std::lock_guard<std::mutex> lock(parent_.handler_mutex_);
-  auto it = parent_.handlers_.find(msg->get_topic());
-  if (it != parent_.handlers_.end())
+  // Copy the handler out and release the lock before invoking it, so a handler
+  // that calls back into subscribe()/unsubscribe() cannot self-deadlock on
+  // handler_mutex_.
+  MqttClientInterface::MessageHandler handler;
   {
-    it->second(msg->get_topic(), msg->get_payload());
+    std::lock_guard<std::mutex> lock(parent_.handler_mutex_);
+    auto it = parent_.handlers_.find(msg->get_topic());
+    if (it == parent_.handlers_.end()) return;
+    handler = it->second;
   }
+  handler(msg->get_topic(), msg->get_payload());
 }
 
 //=============================================================================
@@ -159,7 +164,11 @@ void PahoMqttClient::disconnect()
 
   try
   {
-    conn_options_.set_automatic_reconnect(false);
+    // Note: do not mutate conn_options_ here. Disabling automatic reconnect on
+    // the persistent member has no effect on the already-established session
+    // (Paho captured the options at connect time and stops reconnecting on an
+    // explicit disconnect anyway), but it would corrupt the options reused by
+    // the next connect(), permanently disabling reconnection.
     const auto timeout_ms = static_cast<int>(operation_timeout_.count());
     mqtt::disconnect_options options(timeout_ms);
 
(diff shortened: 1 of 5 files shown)


Briancbn:
Thanks @GameTL for the contribution. I agree that the API should allow us to set custom timeout for the request. We should include this in the API. Let's proceed with the review.

Regarding the issue that you are facing.

> `disconnect()` returns promptly: it short-circuits when not connected and passes the timeout through to Paho's `disconnect_options`.

I reverted the changes made in the `PahoMqttClient` class and retained the [`mqtt_stay_connected.cpp` test script](https://github.com/Briancbn/vda5050_client/blob/feat/mqtt-operation-timeout-example-test/vda5050_core/examples/mqtt_stay_connected.cpp). I was not able to reproduce the timeout error mentioned here. Here are versions I was testing against

Humble:
- Ubuntu 22.04
- mosquitto broker from `eclipse-mosquitto:2` docker image
- `libpaho-mqtt-dev` (`1.3.9-1build1`), `libpaho-mqttpp-dev` (`1.2.0-2`).

Jazzy:
- Ubuntu 24.04
- mosquitto broker from `eclipse-mosquitto:2` docker image
- `libpaho-mqtt-dev` (`1.3.13-1build2`), `libpaho-mqttpp-dev` (`1.2.0-2`).

> Asserts connect/disconnect to an unreachable broker completes within the configured timeout rather than hanging.

I tried with both the mosquitto broker on and off.

I suspect this might be due to https://github.com/eclipse-paho/paho.mqtt.cpp/issues/499 (which is fixed in `paho-mqtt` 1.3.14). I checked your example when implementing the `Adapter`, there does seem to be [mqtt calls potentially coming from multiple threads](https://github.com/GameTL/vda5050_core/blob/d3066aee80342f179c8e53878caeed2ae37cc4c7/vda5050_core/examples/adapter_example.cpp#L124-L128).

vda5050_core/src/transport/paho_mqtt_client.cpp
```diff
@@ -115,14 +120,35 @@ std::unique_ptr<PahoMqttClient> PahoMqttClient::make_unique(
 //=============================================================================
 PahoMqttClient::~PahoMqttClient() = default;
 
+//=============================================================================
+void PahoMqttClient::set_operation_timeout(std::chrono::milliseconds timeout)
+{
+  operation_timeout_ = timeout;
```
sauk2, line 126: If `set_operation_timeout()` is called after the client is in use then we will have to make access to it in `wait_for_token()` be thread-safe, since it can be read concurrently by connect/publish/subscribe operations.

vda5050_core/src/transport/paho_mqtt_client.cpp
```diff
@@ -167,7 +202,7 @@ void PahoMqttClient::publish(
     msg->set_qos(qos);
     msg->set_retained(retain);
 
-    client_->publish(msg)->wait();
+    wait_for_token(client_->publish(msg), "publish");
```
sauk2, line 205: With the introduction of waits, operations can now time out without throwing. Have we considered whether the interface should expose timeout/failure information to callers? Currently a timed-out publish/subscribe may appear successful from the caller's perspective.

sauk2: Not necessarily a blocker but I think worth discussing

vda5050_core/src/transport/paho_mqtt_client.cpp
```diff
@@ -80,12 +80,17 @@ void MqttCallback::connection_lost(const std::string& /*cause*/)
...
+  // handler_mutex_.
+  MqttClientInterface::MessageHandler handler;
   {
-    it->second(msg->get_topic(), msg->get_payload());
+    std::lock_guard<std::mutex> lock(parent_.handler_mutex_);
+    auto it = parent_.handlers_.find(msg->get_topic());
+    if (it == parent_.handlers_.end()) return;
+    handler = it->second;
   }
+  handler(msg->get_topic(), msg->get_payload());
```
sauk2, line 93: Should handler invocation be wrapped in a try/catch block? A user-provided callback could potentially throw an exception. 

This was probably missed in previous reviews

vda5050_core/src/transport/paho_mqtt_client.cpp
```diff
@@ -179,14 +214,27 @@ void PahoMqttClient::publish(
...
   }
+
+  try
+  {
+    if (!wait_for_token(client_->subscribe(topic, qos), "subscribe"))
+    {
+      // Subscription did not complete; roll back so local state stays
+      // consistent with the broker.
+      std::lock_guard<std::mutex> lock(handler_mutex_);
+      handlers_.erase(topic);
```
sauk2, line 231: For subscribe, timeout does not necessarily imply the broker rejected the subscription. Could rolling back the local handler on timeout leave local state inconsistent if the broker later processes the SUBSCRIBE successfully? It may be worth distinguishing timeout from explicit failure.

For example, a possible failure sequence could be,
1. Handler registered locally
2. `SUBSCRIBE` token sent
3. Broker accepts subscription
4. Network stalls
5. `wait_for_token()` times out
6. Local handler erased
7. Broker still considers us subscribed

vda5050_core/src/transport/paho_mqtt_client.cpp
```diff
@@ -115,14 +120,35 @@ std::unique_ptr<PahoMqttClient> PahoMqttClient::make_unique(
...
 //=============================================================================
 void PahoMqttClient::connect()
 {
   if (client_->is_connected()) return;
 
   try
   {
-    client_->connect(conn_options_, nullptr, action_listener_)->wait();
+    wait_for_token(
+      client_->connect(conn_options_, nullptr, action_listener_), "connect");
```
sauk2, line 151: Currently we have automatic reconnect enabled, so the semantics have become a little misleading. 



According to the current Paho implementation,
- `connect()` triggers a connection attempt where success is eventual and state may change asynchronously
- but the background thread may still be retrying connection or reconnecting after network loss
- this means the connection state can change after the function already returned
- so a `connect()` timeout no longer cleanly maps to a failed connection attempt. The client may still become connected shortly after the timeout due to background reconnection.

So the following pattern becomes unreliable,

```cpp
connect();
if (connected())
```

I don’t think we necessarily need to change the implementation, but it would help a lot to do the following,
- Clarify in documentation that `connect()` is a best-effort time-bound attempt, not a guaranteed connection
- Clarify that with auto-reconnect enabled, connection state is inherently asynchronous and may change after `connect()` returns.
- Consider whether code that requires deterministic startup should instead rely on a method that explicitly blocks until connected

sauk2 (changes requested):
Thanks for the PR! This was definitely a necessary change. The improvements to timeout handling and overall hardening of the MQTT transport layer make the system significantly more robust, especially in failure scenarios like unreachable brokers.

Nice work on hardening this layer without changing the public interface!

I have provided my initial comments. @Briancbn also feel free to comment on the concerns I raised.

vda5050_core/src/transport/paho_mqtt_client.cpp
Briancbn: We can simply document this that this is not thread safe?

sauk2: That works!

sauk2:
For now we will work on the features needed for release before we come fix this.

PR #62, opened 2026-06-03
feat(client): add AGVContext and execution tracking resource

eileentyz:
## Summary
Implements the concrete AGV-side context for managing VDA5050 execution resources and state.

### Resources
* **OrderExecutionResource** — Holds the AGV's active execution flags (executing_order, awaiting_order_update) and a vda5050_core::types::State snapshot, protected by its own internal mutex.

### Context
* **AGVContext : ContextInterface** — on `init()`, seeds the two resources and registers `provider()->on<OrderUpdate>()` so each inbound order is cached and fires `notify_on_change()` to wake the Handler.

* Type-indexed update/resource maps behind `get_update<T>()`/`get_resource<T>()`, guarded by a single storage mutex.

### Tests
* **AGVContextTest** — 10 tests: resource seeding, update caching, callback triggering and context registry behaviour.
* **OrderExecutionResourceTest** — resource snapshot storage and distinct type-index identity.


vda5050_core/include/vda5050_core/client/resources/order_execution.hpp
```diff
@@ -0,0 +1,60 @@
...
+/// \brief Persistent, mutable AGV execution state shared across strategies.
+struct OrderExecutionResource
+: public execution::Initialize<OrderExecutionResource, execution::ResourceBase>
+{
+  /// \brief Serialises concurrent reads and writes to this resource.
+  ///
+  /// Callers that reach this resource via get_resource<OrderExecutionResource>()
+  /// must hold `mutex_` while reading or writing `state` (and the flags below),
+  /// as multiple strategies may access it concurrently.
+  mutable std::mutex mutex_;
```
sauk2: I don't think it is a good idea to expose a mutex and rely on callers to manually acquire it before accessing the data. I would consider hiding the mutex and encapsulating the access behind some helper functions.

vda5050_core/include/vda5050_core/client/contexts/agv_context.hpp
sauk2, line 1: I would suggest moving this file under `client/contexts/`

vda5050_core/include/vda5050_core/client/updates/order_context.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+/// \brief AGV client context: in-memory storage for VDA5050 execution data.
+///
+/// Seeds `HeaderConfigResource` and `OrderExecutionResource` as persistent
+/// resources. Registers a provider callback for `OrderUpdate` so strategies
+/// can read the latest inbound order via get_update<OrderUpdate>().
+///
+/// All map reads and writes are protected by a single `storage_mutex_` via
+/// std::lock_guard to prevent race conditions when strategies run concurrently.
+class OrderContext : public execution::ContextInterface
```
sauk2: I have one question related to the design. Since the `Handler` currently owns a single instance of `ContextInterface`, does it make sense to keep it so order specific?

`OrderContext` now exposes a lot of order execution specific methods. I could imagine future implementations needing to work with a lot of additional data this is not directly related to order execution such as actions, connection, state, or other AGV related stuff. 

So, my suggestion would be to keep the context generic and treat order execution as one of the many resources managed by it. In this case, something like `AGVContext` might better reflect its role as a shared execution context for the entire client rather than only order handling.

sauk2 (changes requested):
@eileentyz Thank you for the PR! I have provided my initial comments

vda5050_core/src/client/contexts/agv_context.cpp
```diff
@@ -0,0 +1,82 @@
...
+void AGVContext::init()
+{
+  cache_resource(config_);
+  cache_resource(execution_);
+
+  // Register provider callbacks so incoming updates are cached automatically
+  provider()->on<OrderUpdate>([this](std::shared_ptr<OrderUpdate> update) {
+    this->cache_update(update);
+  });
+}
```
sauk2, line 43: I think this function probably needs a guard of some sort in case it gets called twice. Maybe some store attribute 

```cpp
bool initialized_ = false;
```

vda5050_core/include/vda5050_core/client/resources/order_execution.hpp
```diff
@@ -0,0 +1,173 @@
...
+    state_ = std::move(state);
+  }
+
+  /// \brief Mutate the working state under the resource lock.
+  template <typename Fn>
+  void update_state(Fn&& fn)
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    fn(state_);
+  }
```
sauk2: This function seems a bit risky as it is very easy to misuse. For example, if someone does the following

```cpp
resource->update_state([&] (auto& state) {
  if (resource->is_executing_order())
  {
      // .....
      // .....
  }
});
```
it will result in a deadlock.

Is there a clear use case for exposing a callback-based mutation API? Otherwise, I would suggest a more setter-based API to be safer.

vda5050_core/include/vda5050_core/client/resources/order_execution.hpp
```diff
@@ -0,0 +1,173 @@
...
+/// Owns the working VDA5050 `State` snapshot together with runtime flags for
+/// the current order. Access is synchronised internally, so callers do not
+/// need to manage locking themselves.
+///
+/// Read APIs return copies, while `update_state()` allows safe in-place
+/// modification under the internal mutex.
+///
+/// Multiple strategies may access this resource concurrently via
+/// get_resource<OrderExecutionResource>().
+class OrderExecutionResource
```
sauk2, line 33: I am wondering if we need so many getters. Because we are already exposing the state through `get_state()`, would it be better to let the users extract themselves from the `State`?

vda5050_core/src/client/contexts/agv_context.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+void AGVContext::init()
+{
+  cache_resource(config_);
+  cache_resource(execution_);
+
+  // Register provider callbacks so incoming updates are cached automatically
+  provider()->on<OrderUpdate>([this](std::shared_ptr<OrderUpdate> update) {
+    this->cache_update(update);
+  });
```
sauk2: This callback captures `this`, but `Provider` is exposed externally as a `shared_ptr` and may outlive the context. If a caller stores the provider after context has been destroyed, the callback will try to invoke `cache_update()` on a dangling pointer. I would suggest passing in a `weak_ptr` instead

```suggestion
  provider()->on<OrderUpdate>([w = weak_from_this()](std::shared_ptr<OrderUpdate> update) {
  if (auto con = w.lock())
  {
     con->cache_update(update);
  }
  });
```

However, for this to work, you will have to first make sure `AGVContext` derives from `enable_shared_from_this`

```cpp
class AGVContext : public execution::ContextInterface, public std::enable_shared_from_this<AGVContext>
```

Then maybe we should also hide the constructor and provide a function that returns a `shared_ptr` so that this class cannot be created as is

```cpp
public:

static std::shared_ptr<AGVContext> make()
{
  auto context = std::shared_ptr<AGVContext>(new AGVContext());
  return context;
}
```

vda5050_core/include/vda5050_core/client/contexts/agv_context.hpp
```diff
@@ -0,0 +1,105 @@
...
+class AGVContext : public execution::ContextInterface
+{
+public:
+  /// \brief Construct with AGV identity resource.
+  ///
+  /// `HeaderConfigResource` and a blank `OrderExecutionResource` are available
+  /// via get_resource<T>() after init() is called.
+  explicit AGVContext(std::shared_ptr<HeaderConfigResource> config)
+  : config_(std::move(config)),
+    execution_(std::make_shared<OrderExecutionResource>())
```
sauk2: I see that `config_` and `execution_` are stored both as member variables and also inserted into the resource registry. Instead of later pushing them into the registry through `init()`, would it make sense to register these resources directly during construction and avoid keeping them as separate attributes?

sauk2 (changes requested):
Thanks for the changes! I have provided feedback on the things I missed last review

sauk2 (approved):

PR #63, opened 2026-06-04
feat(client): add order validator and acceptance strategy

eileentyz:
> **Stacked PR** — depends on https://github.com/ros-industrial/vda5050_core/pull/62
Please review / merge in order: https://github.com/ros-industrial/vda5050_core/pull/62 → this PR.

## Summary
Implements the AGV-side order acceptance flow. Processes inbound orders from the `OrderContext`, runs the accept/reject/ignore decision and applies the result to the execution state.

### Strategies
* **OrderValidator** — Handles the validation decision and returns `AcceptanceResult {outcome, errors}` where outcome is one of the `{Accepted, Rejected, Ignored}`:
    - **Structural validity** — validates the order graph structure (Node -> Edge -> Node -> Edge)
    - **New vs Update order detection** — compares `orderId` against the active order
    - **Busy/horizon check** — rejects a new order while the vehicle is still executing or holding non-finished/failed actions
    - **Deprecated update** — lower `orderUpdateId` -> `orderUpdateError`, while keeping previous order
    - **Duplicate update** — equal orderUpdateId is ignored without changing execution state
    - **Stitch continuity** — update's first node must match the current decision point / end of the released base
    - **Future checks** — TODOs are left for AGV capability/factsheet validation and start-node reachability once the required data is available
              
* **OrderAcceptance** — Applies the validator decision. `step()` pulls the latest `OrderUpdate` from the context, runs the validator and updates the execution resource:
    - **Accepted (new order)** — populate `node_states`/ `edge_states` / `action_states`, set order ids and mark `executing_order`
    - **Accepted (old order -> update)** — drop the old horizon, append the new states, bump id
    - **Rejected** — record the errors in `state.errors`
    - **Ignored** — no-op for duplicate resend

### Errors
* **error_codes** — adds shared error/reference constants used by the order acceptance flow such as `validationError`, `orderError`

### Tests
* **OrderValidatorTest** — covers structural reject, busy reject, valid stitch update, duplicate update, deprecated update, empty update and stitch mismatch
* **OrderAcceptanceTest** — covers new order state population, invalid order error recording, order update appending, duplicate ignored behaviour

### Notes
- Reachability, capability hooks and protocol limit checks were not added in this PR to keep the scope focused on the core order acceptance flow.
- TODOs were left for future AGV-specific checks once map, position, factsheet and capability context are available.

### DIFF
--- a/vda5050_core/include/vda5050_core/client/strategies/order_acceptance.hpp
+++ b/vda5050_core/include/vda5050_core/client/strategies/order_acceptance.hpp
@@ -47,8 +47,7 @@ class OrderAcceptance : public execution::StrategyInterface
 public:
   OrderAcceptance();
 
-  /// \brief Construct with a pre-configured validator (e.g. with real
-  /// reachability/capability hooks injected by the robot config layer).
+  /// \brief Construct with a pre-configured validator
   explicit OrderAcceptance(OrderValidator validator)
   : validator_(std::move(validator))
   {
--- a/vda5050_core/include/vda5050_core/client/strategies/order_acceptance.hpp
+++ b/vda5050_core/include/vda5050_core/client/strategies/order_acceptance.hpp
@@ -56,9 +56,6 @@ public:
 private:
   OrderValidator validator_;
 
-  // Last (orderId, orderUpdateId) processed. The inbound OrderUpdate stays
-  // cached and step() runs every handler tick, so this lets an unchanged order
-  // be processed once instead of re-validated (and re-logged) on every tick.
   std::string last_order_id_;
   uint32_t last_order_update_id_ = 0;
 };
--- a/vda5050_core/include/vda5050_core/client/strategies/order_acceptance.hpp
+++ b/vda5050_core/include/vda5050_core/client/strategies/order_acceptance.hpp
@@ -22,7 +22,6 @@
 #include <cstdint>
 #include <memory>
 #include <string>
-#include <utility>
 
 #include "vda5050_core/client/strategies/order_validator.hpp"
 #include "vda5050_core/execution/context_interface.hpp"
@@ -48,10 +47,7 @@ public:
   OrderAcceptance();
 
   /// \brief Construct with a pre-configured validator
-  explicit OrderAcceptance(OrderValidator validator)
-  : validator_(std::move(validator))
-  {
-  }
+  explicit OrderAcceptance(OrderValidator validator);
 
   void init(std::shared_ptr<execution::ContextInterface> context) override;
 
--- a/vda5050_core/include/vda5050_core/client/strategies/order_acceptance.hpp
+++ b/vda5050_core/include/vda5050_core/client/strategies/order_acceptance.hpp
@@ -19,7 +19,9 @@
 #ifndef VDA5050_CORE__CLIENT__STRATEGIES__ORDER_ACCEPTANCE_HPP_
 #define VDA5050_CORE__CLIENT__STRATEGIES__ORDER_ACCEPTANCE_HPP_
 
+#include <cstdint>
 #include <memory>
+#include <string>
 #include <utility>
 
 #include "vda5050_core/client/strategies/order_validator.hpp"
@@ -57,6 +59,12 @@ public:
 
 private:
   OrderValidator validator_;
+
+  // Last (orderId, orderUpdateId) processed. The inbound OrderUpdate stays
+  // cached and step() runs every handler tick, so this lets an unchanged order
+  // be processed once instead of re-validated (and re-logged) on every tick.
+  std::string last_order_id_;
+  uint32_t last_order_update_id_ = 0;
 };
 
 }  // namespace client
--- a/vda5050_core/include/vda5050_core/client/strategies/order_acceptance.hpp
+++ b/vda5050_core/include/vda5050_core/client/strategies/order_acceptance.hpp
@@ -1,0 +1,68 @@
+/*
+ * Copyright (C) 2026 ROS-Industrial Consortium Asia Pacific
+ * Advanced Remanufacturing and Technology Centre
+ * A*STAR Research Entities (Co. Registration No. 199702110H)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef VDA5050_CORE__CLIENT__STRATEGIES__ORDER_ACCEPTANCE_HPP_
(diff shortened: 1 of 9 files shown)


vda5050_core/src/client/strategies/order_acceptance.cpp
```diff
@@ -0,0 +1,233 @@
...
+  // orderUpdateId). Without this the validator re-evaluates and logs the same
+  // order ~10x/second for as long as it stays cached.
+  if (
+    order.order_id == last_order_id_ &&
+    order.order_update_id == last_order_update_id_)
+  {
+    return;
+  }
+  last_order_id_ = order.order_id;
+  last_order_update_id_ = order.order_update_id;
```
stevelimyh: I think there may be an unhandled case here.

This stores the last seen `order_id` / `order_update_id` before knowing whether the order was accepted. If a new order is rejected because the AGV is busy, and the master later re-sends the same order once the AGV is free, it will be skipped here and never accepted.

Could we either clear the remembered order when the AGV becomes idle, or only update `last_order_id_` / `last_order_update_id_` after the order is accepted?

vda5050_core/src/client/strategies/order_acceptance.cpp
```diff
@@ -0,0 +1,233 @@
...
+}
+
+void OrderAcceptance::step(std::shared_ptr<execution::ContextInterface> context)
+{
+  auto order_context = std::dynamic_pointer_cast<AGVContext>(context);
+  if (!order_context)
+  {
+    VDA5050_WARN_STREAM("OrderAcceptance: context is not an AGVContext");
+    return;
+  }
```
stevelimyh: `step()` takes a `ContextInterface`, but immediately casts it to the concrete `AGVContext`. Since the methods used here, `get_update<>()` and `get_resource<>()`, are available on `ContextInterface`, could we use the base interface directly?

That would avoid the `dynamic_pointer_cast` and allow the strategy to work with any compatible context. It would also avoid the current failure mode where a wrong context type only produces a repeated warning and return.

vda5050_core/src/client/strategies/order_validator.cpp
```diff
@@ -0,0 +1,228 @@
...
+  }
+
+  // Execution state (order/node tracking) lives in the resource, not the
+  // context itself; the context only routes us to it.
+  auto execution = context->get_resource<OrderExecutionResource>();
+  if (!execution)
+  {
+    return rejected(errors::create_error(
+      errors::ValidationError, "OrderExecutionResource is null", {}));
+  }
```
stevelimyh, line 131: Same point as in `order_acceptance.cpp`: `validate_order()` takes the concrete `AGVContext`, but only uses `get_resource<OrderExecutionResource>()`, which is already available on `ContextInterface`. Could this take `ContextInterface` instead?

vda5050_core/src/client/strategies/order_validator.cpp
```diff
@@ -0,0 +1,228 @@
...
+  {
+    if (
+      action.action_status != types::ActionStatus::FINISHED &&
+      action.action_status != types::ActionStatus::FAILED)
+    {
+      return true;
+    }
+  }
+  return false;
+}
```
stevelimyh, line 112: `is_busy()` treats the AGV as busy whenever `node_states` is non-empty, but this PR does not remove nodes as the AGV reaches them yet.

Once an order is accepted, new orders will continue to be rejected until the traversal/state-update handling lands. Consider adding a short note here to make that limitation explicit.

stevelimyh (changes requested):
@eileentyz Thanks for the PR! I’ve left a few initial comments.

vda5050_core/src/client/strategies/order_acceptance.cpp
eileentyz: Thank you for pointing this out! I have moved the update of `last_order_id_` / `last_order_update_id_` into the `Accepted` case, so rejected orders are no longer remembered as processed.

I also added a test to cover the case where an order is rejected while the AGV is busy, then resent and accepted once the AGV becomes idle.

vda5050_core/src/client/strategies/order_acceptance.cpp
eileentyz: Thank you for pointing this out! I have updated OrderAcceptance to use ContextInterface directly instead of casting to AGVContext.

vda5050_core/src/client/strategies/order_validator.cpp
eileentyz: Thank you for pointing this out! I have also updated OrderValidator to use ContextInterface directly instead of casting to AGVContext.

vda5050_core/src/client/strategies/order_validator.cpp
eileentyz: Thank you for pointing this out! I have added a note to make the current limitation explicit.

stevelimyh (approved):
Thanks for the changes @eileentyz, LGTM. @sauk2 could you take a look?

vda5050_core/include/vda5050_core/client/strategies/order_validator.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+/// \brief Outcome of the VDA5050 order-acceptance decision.
+enum class AcceptanceOutcome
+{
+  /// \brief Order is valid and should be applied to the execution state.
+  Accepted,
+  /// \brief Order is invalid; report `errors` and keep the previous order.
+  Rejected,
+  /// \brief Duplicate resend (same orderId + orderUpdateId); leave state unchanged.
+  Ignored
```
sauk2:
```suggestion
  ACCEPTED,
  /// \brief Order is invalid; report `errors` and keep the previous order.
  REJECTED,
  /// \brief Duplicate resend (same orderId + orderUpdateId); leave state unchanged.
  IGNORED
```

vda5050_core/include/vda5050_core/client/strategies/order_acceptance.hpp
```diff
@@ -0,0 +1,73 @@
...
+class OrderAcceptance : public execution::StrategyInterface
+{
+public:
+  OrderAcceptance();
+
+  /// \brief Construct with a pre-configured validator
+  explicit OrderAcceptance(OrderValidator validator)
+  : validator_(std::move(validator))
+  {
+  }
```
sauk2: Move this to `.cpp`

vda5050_core/src/client/strategies/order_acceptance.cpp
```diff
@@ -0,0 +1,221 @@
...
+    std::remove_if(
+      state.edge_states.begin(), state.edge_states.end(),
+      [](const types::EdgeState& e) { return !e.released; }),
+    state.edge_states.end());
+
+  for (size_t i = 1; i < order.nodes.size(); ++i)
+  {
+    append_node(state, order.nodes[i]);
+  }
+  for (const auto& edge : order.edges) append_edge(state, edge);
```
sauk2: Why are we using two different ways to append objects to vector?

sauk2 (approved):
Thanks for this PR! Just some minor comments

vda5050_core/src/client/strategies/order_acceptance.cpp
eileentyz: Thank you for pointing this out! I have updated both loops to use the same index-based style for consistency.

PR #64, opened 2026-06-04
feat(client): add order traversal strategy

eileentyz:
> **Stacked PR** — depends on https://github.com/ros-industrial/vda5050_core/pull/63
> Please review / merge in order: https://github.com/ros-industrial/vda5050_core/pull/62 → https://github.com/ros-industrial/vda5050_core/pull/63 → this PR.

## Summary
Implements the AGV-side order traversal strategy for walking the released base of an accepted order.

While an order is executing, the strategy checks traversal progress against the reported `types::State` on every step. It applies each `NodeReachedUpdate` at most once, removes the reached `NodeState` and its incoming `EdgeState`, updates `last_node_id` / `last_node_sequence_id`, refreshes `new_base_request`, and selects the next node to traverse toward.

`types::State` remains the source for traversal progress and VDA5050 state reporting. The accepted full `types::Order` is retained separately as the source for execution data. When dispatching navigation, traversal looks up the selected full `types::Node` and optional incoming full `types::Edge` from the stored order, then emits a `NavigateToNodeEvent` carrying `adapter::NodeRequest` and optional `adapter::EdgeRequest`.

This avoids reconstructing a full edge from `EdgeState` qhich is only a reporting view and does not contain all execution fields such as `max_speed`, `height limits`, etc.

Because the strategy checks the released base every step rather than only when a node is reached, the first released node can be dispatched on order start, and a base extension can resume traversal from a decision point without requiring a new `NodeReachedUpdate`.

### Event
* **NavigateToNodeEvent** — emitted through the Engine for the navigation layer to consume. Carries an `adapter::NodeRequest` target and when available, an `adapter::EdgeRequest` representing the incoming edge used to reach it. `via_edge` is empty when no incoming edge is available.

### Inbound signal
* **NodeReachedUpdate** — signal pushed by the vehicle/localisation layer when the vehicle reaches a node. Carries `node_id` and `sequence_id`.

### Strategy
* **OrderTraversal** — while an order is executing (`is_executing_order()`), each `step()` checks traversal against the released base:
    - uses `types::State` for traversal progress, pending `node_states` / `edge_states`, `last_node_id`, `last_node_sequence_id`, and `new_base_request`
    - applies the latest `NodeReachedUpdate` at most once
    - ignores the update unless the reported node is the next expected one
    - updates `last_node_id` / `last_node_sequence_id`
    - removes the reached node's `NodeState`
    - removes the reached node's incoming `EdgeState`
    - refreshes `new_base_request` based on the released base ahead
    - selects the next node by the smallest `sequenceId` strictly ahead of `last_node_sequence_id`
    - stops at the decision point when the next node is horizon/unreleased or when there is no node ahead
    - looks up the full selected `types::Node` and optional incoming `types::Edge` from the stored accepted `types::Order`
    - emits `NavigateToNodeEvent` when the next node is released
    - calls `engine()->step()` to deliver the emitted event

The strategy also keeps small local idempotency guards to avoid re-processing the same cached `NodeReachedUpdate` and to avoid re-emitting the same navigation target repeatedly for the same order.

### Tests
* **OrderTraversalTest** — covers advances state on node reached, ignores an unknown node-reached signal, stops at the decision point and raises new_base_request...

## Not Included
  * Action execution is not implemented in this PR. A later action strategy will use the accepted full order for node/edge action data.


vda5050_core/src/client/strategies/order_traversal.cpp
```diff
@@ -0,0 +1,210 @@
...
+// Finds the next node after the last reached node.
+// Returns nullopt if there is no node ahead or the next node is still horizon.
+std::optional<Dispatch> compute_next_dispatch(const types::State& state)
+{
+  const types::NodeState* next = nullptr;
+  for (const auto& node : state.node_states)
+  {
+    if (node.sequence_id <= state.last_node_sequence_id) continue;
+    if (next == nullptr || node.sequence_id < next->sequence_id) next = &node;
+  }
```
sauk2: This loop is O(N) on every step. Instead, can we cache a pointer somewhere to reduce the number of iterations

vda5050_core/src/client/strategies/order_traversal.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+void OrderTraversal::init(
+  std::shared_ptr<execution::ContextInterface> /*context*/)
+{
+  // Nothing to initialise; node-reached signals are read from the context.
+}
+
+void OrderTraversal::step(std::shared_ptr<execution::ContextInterface> context)
+{
+  auto order_context = std::dynamic_pointer_cast<AGVContext>(context);
```
sauk2: I don't think you need to cast the context here. The generic `ContextInterface::get_resource` should be sufficient to access any data you need

vda5050_core/include/vda5050_core/client/events/navigate_to_node.hpp
```diff
@@ -0,0 +1,60 @@
...
+struct NavigateToNodeEvent
+: public execution::Initialize<NavigateToNodeEvent, execution::EventBase>
+{
+  /// \brief Node that the AGV should navigate to.
+  types::NodeState target;
+
+  /// \brief Edge to traverse before reaching the target node.
+  ///
+  /// Empty when navigating to the first node of the order.
+  std::optional<types::EdgeState> via_edge;
```
sauk2: One thing to note here is that we should explicitly control how navigation commands and actions are dispatched. Right now, this event bundles the next node/edge along with any associated actions, which would push that responsibility to the consumer to interpret blocking behavior and execute actions correctly.

Instead, the adapter API is intentionally split into `on_navigate` and `on_action` to separate concerns and allow us to stagger execution - https://github.com/sauk2/vda5050_core/blob/21a150498e1a6ed270864d43485367e07502bec2/vda5050_core/include/vda5050_core/client/adapter/adapter.hpp#L53-L60

As you can see, `NodeRequest` and `EdgeRequest` are deliberately kept small without the exposure of actions, so that action execution is handled independently through the `on_action` callback.

I would suggest you to align this event with the adapter design.

sauk2 (changes requested):
Thanks for the great work! I have provided my initial comments

vda5050_core/include/vda5050_core/client/events/navigate_to_node.hpp
eileentyz: Thanks for pointing this out! I added the adapter’s `NodeRequest` and `EdgeRequest` files to this branch and updated `NavigateToNodeEvent` to use these request types directly.

However, I made the `from_node()` and `from_edge()` public so that `OrderTraversal` can construct the requests. I also updated `OrderExecutionResource` and `OrderAcceptance` to retain the full accepted order. Previously, traversal reconstructed the node and edge from `NodeState` and `EdgeState`, which could result in navigation fields such as max_speed and height_limits being lost because they are not included in the state representation.

`types::State` now is still used for traversal progress and state reporting, while the retained full order is used to construct complete navigation requests. Please let me know if you would prefer a different design approach!

PR #69, opened 2026-06-17
feat(master): add order lifecycle, stitcher, instant actions, mode handling, and pose_view

stevelimyh:
## Summary

Adds the master-logic layer and public C++ API for the VDA5050 master restructure.

This layer sits between the validator chain and the upcoming ROS 2 API layer. It exposes `VDA5050Master`, the abstract base class that integrators subclass to command and observe a fleet. It tracks dispatched orders, handles order updates through stitching, dispatches instant actions, reacts to operating-mode changes, and exposes a fused per-AGV pose view.

This PR builds on the `layout::Graph`-based validators (#55, #56) and the fleet-wide `MasterContext` update detector (#58), which are now merged into `develop`, so this PR targets `develop` directly.

## What's included

- **C++ master API** — `VDA5050Master`, the abstract base class for fleet integration:
  - connect / disconnect
  - onboard / offboard, including batch onboarding
  - `assign_order` / `assign_instant_actions`
  - topology loading
  - broker status
  - `on_*` observation and edge-trigger hooks for node reached, errors, mode changes, state timeout, broker connect/disconnect, and related master events
  - API documentation in `vda5050_core/docs/master_api.md`

- **Order management** — `OrderLifecycleManager` for per-AGV active-order tracking, pending-update queue, completion state, and mismatch/handover recovery; `OrderStitcher` for base/update merging; and `OrderPublisher` for outbound dispatch.

- **Instant actions** — `ActionFactory` typed builders, `InstantActionsPublisher` with shared validation gate, and assignment-result plumbing for instant-action dispatch.

- **State and mode handling** — pending-queue capture/drain across operating-mode changes, state-silence timeout detection, and broker disconnect/reconnect observability hooks.

- **Pose view** — `PoseView` and `AGV::get_pose_view()` for fused `State` / `Visualization` pose data. The ROS 2 `pose_view` topic lands in the next PR; the `GetPoseView` service is deferred.

## Integration

`agv.cpp` and `master.cpp` wire lifecycle management, stitching, action dispatch, mode handling, timeout handling, broker-status handling, and pose view into the per-AGV state machine, surfaced through the `VDA5050Master` API.

The master also loads the topology layout as a `layout::Graph` and refreshes per-AGV factsheet-alignment data on layout load and factsheet arrival.

## Testing

Added unit and integration tests covering order lifecycle/stitching, order and instant-action dispatch, mode capture/drain, state timeout detection, broker-status handling, pose view, configurable interface name, and onboard batch flows.


vda5050_core/src/master/master.cpp
```diff
@@ -57,11 +119,20 @@ void VDA5050Master::connect()
     return;
   }
 
+  // Capture weak_ptr, not `this`: lock() no-ops once the master is gone, and
+  // keeps it alive while held (callbacks can race destruction).
+  std::weak_ptr<VDA5050Master> weak = weak_from_this();
```
sauk2, line 144: Because the constructor of `VDA5050Master` is public, there is no guarantee that the user will create it using `std::make_shared`. I would suggest making the constructor private and exposing a function so that users cannot freely construct or inherit this class

```cpp
static std::shared_ptr<VDA5050Master> make(...);
```

vda5050_core/include/vda5050_core/master/master.hpp
```diff
@@ -19,263 +19,453 @@
...
 
-  /**
-   * @brief Get a shared pointer to an onboarded AGV
-   * @param manufacturer Manufacturer name
-   * @param serial_number Serial number
-   * @return Shared pointer to AGV, or nullptr if not onboarded
-   */
+  /// \brief The onboarded AGV, or nullptr if not onboarded.
   std::shared_ptr<AGV> get_agv(
     const std::string& manufacturer, const std::string& serial_number) const;
```
sauk2: I just noticed this function could cause a problem. Since functions like `send_order`, `send_instant_actions` etc are public, the user could bypass the master and call them instead. 

So my suggestion would be to mark it as a const pointer so only the getters are accessible 

```suggestion
std::shared_ptr<const AGV> get_agv(
const std::string& manufacturer, const std::string& serial_number) const;
```

vda5050_core/src/master/master.cpp
```diff
@@ -200,17 +583,128 @@ bool VDA5050Master::publish_order(
...
+      res.decision = AssignmentDecision::DUPLICATE_IGNORED;
+      return res;
+    }
+    stitch_will_queue = (stitch.decision == StitchDecision::QUEUE_PENDING);
+  }
+
+  // send_order returns false only when the outbound queue is full.
+  if (!agv->send_order(order))
+  {
+    res.decision = AssignmentDecision::AGV_NOT_READY;
```
sauk2: This looks like a wrong result for a full queue. I see that instant actions has an enum to indicate a full queue. Can we have the same here?

vda5050_core/include/vda5050_core/master/master.hpp
```diff
@@ -19,263 +19,453 @@
...
+  void record_assignment(
+    const std::string& manufacturer, const std::string& serial_number,
+    const std::string& assignment_id, const std::string& order_id,
+    std::uint32_t order_update_id);
+
+  std::string get_active_assignment_id(
+    const std::string& manufacturer, const std::string& serial_number) const;
+
+  void clear_assignment(
+    const std::string& manufacturer, const std::string& serial_number);
```
sauk2: Should these be private instead? They seem to be used internally and I don't think they should be exposed to the user

vda5050_core/include/vda5050_core/master/agv.hpp
```diff
@@ -411,91 +303,123 @@ class AGV : public std::enable_shared_from_this<AGV>
...
+  OrderStitcher order_stitcher_;
+
+  // Set once at construction, never reassigned (safe to read concurrently);
+  // weak_ptr so dispatch detects master destruction via lock().
   std::weak_ptr<VDA5050Master> parent_;
 
-  // Heartbeat listener for state timeout detection (protected by heartbeat_mutex_)
+  // Used ONLY by the queue-processor thread (parent_.lock() there could
+  // self-join ~master->~AGV); master joins AGV threads first, stays valid.
+  VDA5050Master* parent_raw_{nullptr};
```
sauk2: Can you use the `parent_` variable instead? A weak ptr looks safer to use

sauk2 (approved):
Thanks for the API redesign. Just a few changes and then this can be merged!

vda5050_core/src/master/master.cpp
stevelimyh: Thanks for the review!
Addressed in 16ebc37

vda5050_core/include/vda5050_core/master/master.hpp
stevelimyh: Done in 2db0aee. `get_agv()` now returns `std::shared_ptr<const AGV>`, so callers can only access the read-only AGV surface. Since that also hides write paths like `cancel_pending_orders()` and mode-cancelled resume/discard, those operations were moved onto the master.

vda5050_core/src/master/master.cpp
stevelimyh: Done in 0ea302a

vda5050_core/include/vda5050_core/master/agv.hpp
stevelimyh: Thanks, agreed the raw pointer is not ideal. I tried `parent_` first, but `lock()` on the queue thread can make that thread the master's last owner. In that case, `~VDA5050Master` would run on the queue thread and try to join that same thread, which terminates.

So instead of swapping the pointer type, I removed the back-reference entirely in fa25437: the loaded graph now lives in a small shared holder held by both the master and each AGV, and the queue thread reads it with no reference to the master at all. That removes both the dangling raw pointer risk and the self-join risk.

vda5050_core/include/vda5050_core/master/master.hpp
stevelimyh: Good point. The `assignment_id` correlation turned out to be unused and redundant with `orderId`, so I removed the feature entirely in e972e4f instead of just hiding those methods.

vda5050_core/CMakeLists.txt
```diff
@@ -96,7 +96,7 @@ target_include_directories(vda5050_errors
 )
 
 # Validation
-file(GLOB_RECURSE validation_srcs "src/validation/*.cpp")
+file(GLOB_RECURSE validation_srcs CONFIGURE_DEPENDS "src/validation/*.cpp")
```
Briancbn: Is this needed?

Briancbn (commented):
Thanks for the first working version of our master API.

I didn't have time to do a detailed review, but from the look, it is still very raw and as we are creeping close to our 0.1.0 release. I propose we keep this under an `experimental` namespace, (under an experimental folder). Let me know your thoughts on this. @sauk2

vda5050_core/include/vda5050_core/master/agv.hpp
sauk2: Okay, then it should be fine for now

sauk2:
> I didn't have time to do a detailed review, but from the look, it is still very raw and as we are creeping close to our 0.1.0 release. I propose we keep this under an `experimental` namespace, (under an experimental folder). 

I would prefer to have a big warning in the documentation instead of adding a new namespace. Also, we can add warning as logs that show up when a user launches the master. What do you think?

vda5050_core/CMakeLists.txt
stevelimyh: Not needed, removed. Also dropped it from master_srcs for consistency.

stevelimyh:
> > I didn't have time to do a detailed review, but from the look, it is still very raw and as we are creeping close to our 0.1.0 release. I propose we keep this under an `experimental` namespace, (under an experimental folder).
> 
> I would prefer to have a big warning in the documentation instead of adding a new namespace. Also, we can add warning as logs that show up when a user launches the master. What do you think?

Agreed that the master API needs a cleanup, likely in a follow-up PR after the `0.1.0` release. I prefer adding a warning over introducing a new namespace for now.

Added a construction-time warning in `3a82b77`; the master API docs will include the same warning as well.

PR #71, opened 2026-06-26
feat(client): provide initial adapter interface along with rmf migration guide

sauk2:
## Summary

This PR provides an initial adapter with a set of APIs to allow users to make a robot VDA5050 compliant with minimal effort. The APIs are in C++ with a set of python bindings provided specifically for migration from Open-RMF fleet adapters. (Note that the C++ APIs do not have direct python bindings yet). A C++ example can be found in `examples/client/adapter_example.cpp` along with an order publisher for testing.

### Running the example

The following code launches a minimal example implementation of the adapter.
```bash
ros2 run vda5050_core adapter_example
```

To test it, start the order publisher the following way.
```bash
ros2 run vda5050_core order_publisher
```
The publisher first sends a series of actions - `factsheetRequest`, `stateRequest` and `initPosition`. Once the AGV is calibrated, it sends an order. The order updates are published as long as the publisher is alive. Kill it to stop the publish of orders.

## Pending tasks

- [ ] Unit tests
- [ ] Integration tests
- [ ] Python adapter example
- [ ] Documentation and RMF migration guide

cc @eileentyz


Briancbn:
For testing purpose, maybe can give some of the adapters in [awesome fleet adapter list](https://github.com/open-rmf/awesome_adapters#fleet-adapters) a try

sauk2:
The DCO kept failing so I had to squash all commits into one and force push. For reviewers who want to see the commit history before https://github.com/ros-industrial/vda5050_core/pull/71/commits/aeba1cbb6e2d526839904c68e87b9dfe6d5a2766, you can check the backup branch for reference - https://github.com/sauk2/vda5050_core/tree/experimental-adapter-backup

vda5050_core/src/client/adapter/adapter.cpp
```diff
@@ -0,0 +1,730 @@
...
+  protocol_adapter->publish<types::Factsheet>(manager->factsheet.value(), 0);
+}
+
+//=============================================================================
+void Adapter::Implementation::publish_state()
+{
+  protocol_adapter->publish<types::State>(state_manager->state(), 0);
+}
+
+//=============================================================================
```
stevelimyh, line 437: I think these two paths can race. Both call `ProtocolAdapter::publish()`, which performs `header_ids_[type_idx]++` on a plain `std::unordered_map` without synchronization.

`publish_factsheet()` runs on the dispatch thread, while `publish_state()` runs on the state thread, so they may access `header_ids_` concurrently when a `factsheetRequest` arrives during a state publication.

`ProtocolAdapter` already protects `active_flags_` with a mutex. Would it make sense to do the same for `header_ids_`?

vda5050_core/src/client/adapter/adapter.cpp
```diff
@@ -0,0 +1,730 @@
...
+    VDA5050_ERROR(
+      "Navigation failed towards node ID [{}] with sequence ID [{}]",
+      next_node.node_id, next_node.sequence_id);
+  }
+}
+
+//=============================================================================
+void Adapter::Implementation::process_actions()
+{
+  if (!action_callback) return;
```
stevelimyh: It looks like this early return happens before the queue is drained, so the built-in `factsheetRequest`, `stateRequest`, and `initPosition` handlers below would not run unless `on_action` is registered, even though they do not seem to depend on it

vda5050_core/src/client/adapter/adapter.cpp
```diff
@@ -0,0 +1,730 @@
...
+    try
+    {
+      action_callback(std::move(request), execution);
+    }
+    catch (const std::exception& e)
+    {
+      VDA5050_ERROR(
+        "Failed to run action of type [{}] and action ID [{}]",
+        action.action_type, action.action_id);
+    }
```
stevelimyh, line 420: I think a malformed `initPosition` action could crash the client here. `action.action_parameters.value()` throws if `actionParameters` is absent, and `std::stod(parameter.value)` throws for an empty or non-numeric value. Both appear to be outside the `try` block in `process_actions`, while the dispatch loop itself has no exception handler, so the exception may escape the thread and call `std::terminate`.

Even though `build_init_position()` would not generate this, maybe we can guard `action_parameters`, validate the parsed values, and catch exceptions before they escape the thread.

vda5050_core/src/client/adapter/adapter.cpp
```diff
@@ -0,0 +1,730 @@
...
+    VDA5050_INFO(
+      "No localization handler discovered. Applying default transformation "
+      "...");
+
+    Pose2D agv_pose{
+      current_position->x, current_position->y, current_position->theta};
+    Transformation tx = Transformation::calibrate(world_pose, agv_pose);
+    state_manager->set_transformation(tx, map_id.value());
+    state_manager->set_last_node(last_node_id.value());
+
```
stevelimyh:
```suggestion
    state_manager->set_last_node(last_node_id.value());
    state_manager->set_position_initialized(true);

```
I think the default transformation path may be missing a step. It calibrates the transform and finishes, but does not appear to call `set_position_initialized(true)` as the `on_localize` path does.

If so, `positionInitialized` would remain false, and the master may continue treating the AGV as unlocalized and hold off sending orders.

vda5050_core/src/client/adapter/adapter.cpp
```diff
@@ -0,0 +1,730 @@
...
+//=============================================================================
+void Adapter::Implementation::subscribe_instant_actions()
+{
+  protocol_adapter->subscribe<types::InstantActions>(
+    [this](types::InstantActions actions, auto error) {
+      if (error.has_value())
+      {
+        state_manager->add_error(error.value());
+        return;
```
stevelimyh, line 158:
```suggestion
      if (error.has_value())
      {
        state_manager->add_error(error.value());
        request_state_publish();
        return;
```
The order parse-error path calls `request_state_publish()`, but this one does not.

vda5050_core/src/client/adapter/adapter_impl.hpp
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef CLIENT__ADAPTER__ADAPTER_IMPL_HPP_
+#define CLIENT__ADAPTER__ADAPTER_IMPL_HPP_
```
stevelimyh, line 20:
```suggestion
#ifndef VDA5050_CORE__CLIENT__ADAPTER__ADAPTER_IMPL_HPP_
#define VDA5050_CORE__CLIENT__ADAPTER__ADAPTER_IMPL_HPP_
```

vda5050_core/src/client/adapter/state_manager.cpp
```diff
@@ -0,0 +1,393 @@
...
+  state_.battery_state = battery;
+}
+
+//=============================================================================
+void StateManager::set_operating_mode(types::OperatingMode mode)
+{
+  std::lock_guard<std::mutex> lock(mutex_);
+  state_.operating_mode = mode;
+}
+
```
stevelimyh, line 117: These setters update fields that should trigger a `State` transmission, but only `request_state_publish()` notifies `state_cv`. A change may therefore not publish until the next event or the 30 second heartbeat.

There also seems to be an inconsistency: `set_driving()` sets `publish_requested_`, while `set_operating_mode()` and `add_load()` do not. Could these setters use a shared helper that sets the flag and wakes the publish thread?

vda5050_core/src/client/adapter/adapter.cpp
```diff
@@ -0,0 +1,730 @@
...
+      "...");
+
+    Pose2D agv_pose{
+      current_position->x, current_position->y, current_position->theta};
+    Transformation tx = Transformation::calibrate(world_pose, agv_pose);
+    state_manager->set_transformation(tx, map_id.value());
+    state_manager->set_last_node(last_node_id.value());
+
+    execution->finished();
+  }
```
stevelimyh: I think calling `initPosition` more than once through the default transform path, when no `on_localize` callback is registered, corrupts the calibration.

The first call works because no transform exists yet, so the `agv_pose` passed to `calibrate()` is still the raw pose. On later calls, `state().agv_position` has already been transformed to the world frame by `set_position()`. `calibrate()` then compares world against world instead of AGV against world, effectively baking the previous transform into the new one.

vda5050_core/src/client/adapter/adapter.cpp
```diff
@@ -0,0 +1,730 @@
...
+{
+  protocol_adapter->publish<types::State>(state_manager->state(), 0);
+}
+
+//=============================================================================
+void Adapter::Implementation::request_state_publish()
+{
+  state_manager->mark_publish_requested();
+  state_cv.notify_all();
+}
```
stevelimyh, line 443:
```suggestion
void Adapter::Implementation::request_state_publish()
{
  std::lock_guard<std::mutex> lock(state_mutex);
  state_manager->mark_publish_requested();
  state_cv.notify_all();
}
```
I think the state thread can miss a publish request here. It waits on `state_cv` under `state_mutex`, while the publish flag is set under `StateManager::mutex_`, so a request can be lost and the `State` may not be published until the next 30-second heartbeat

vda5050_core/src/client/adapter/adapter.cpp
```diff
@@ -0,0 +1,730 @@
...
+  pimpl_->start_state_thread();
+}
+
+//=============================================================================
+void Adapter::stop()
+{
+  if (!pimpl_->running) return;
+
+  pimpl_->running = false;
+  pimpl_->state_cv.notify_all();
```
stevelimyh, line 714:
```suggestion
  {
    std::lock_guard<std::mutex> lock(pimpl_->state_mutex);
    pimpl_->running = false;
  }
  pimpl_->state_cv.notify_all();
```
Same issue on shutdown: without `state_mutex`, the notify may be missed and `stop()` can block for up to 30 seconds

vda5050_core/src/client/adapter/state_manager.cpp
```diff
@@ -0,0 +1,393 @@
...
+  {
+    types::NodeState ns;
+    ns.node_id = node.node_id;
+    ns.sequence_id = node.sequence_id;
+    ns.released = node.released;
+    ns.node_description = node.node_description;
+    ns.node_position = node.node_position;
+
+    state_.node_states.push_back(std::move(ns));
+  }
```
stevelimyh: I think this may stall a stitched order when the stitch or anchor node has already been reached before the update arrives.

For an update, `nodes[0]` is the resent stitch node. If the AGV has already reached it, this loop adds it back into `node_states`. However, `process_navigation()` resumes from `last_completed + 2`, so the anchor node is never dispatched again. It therefore seems to remain in `node_states` indefinitely, preventing the order from being considered complete.

`set_order()` currently cannot distinguish a new order from an update. Would it make sense to split the two paths so that, for an update, we preserve the still-pending released states, remove only the unreleased horizon, and append the updated states starting from the second node, skipping the anchor node?

vda5050_core/src/client/adapter/adapter_impl.hpp
```diff
@@ -0,0 +1,172 @@
...
+    std::shared_ptr<ActionExecution> execution);
+
+  types::Factsheet make_default_factsheet();
+};
+
+}  // namespace adapter
+}  // namespace client
+}  // namespace vda5050_core
+
+#endif  // CLIENT__ADAPTER__ADAPTER_IMPL_HPP_
```
stevelimyh, line 174:
```suggestion
#endif  // VDA5050_CORE__CLIENT__ADAPTER__ADAPTER_IMPL_HPP_
```

stevelimyh (changes requested):
Nice work on this! I left a few comments. Let me know what you think

vda5050_core/python/src/rmf_migration.cpp
```diff
@@ -0,0 +1,540 @@
...
+  update_interval_ = value;
+}
+
+//=============================================================================
+std::shared_ptr<RobotUpdateHandle> FleetUpdateHandle::add_robot(
+  std::string name, RobotState initial_state, RobotConfiguration configuration,
+  RobotCallbacks callbacks)
+{
+  auto mqtt_client = transport::create_default_client_unique(
+    configuration_.broker_uri(), configuration_.client_id_prefix());
```
stevelimyh:
```suggestion
  auto mqtt_client = transport::create_default_client_unique(
    configuration_.broker_uri(), configuration_.client_id_prefix() + "_" + name));
```
I think every robot may end up sharing one MQTT client id,so with two or more robots they would keep knocking each other offline

vda5050_core/python/src/rmf_migration.cpp
```diff
@@ -0,0 +1,540 @@
...
+  pos.map_id = std::move(map);
+
+  return pos;
+}
+
+//=============================================================================
+types::BatteryState RobotState::to_battery_state(double soc)
+{
+  types::BatteryState b;
+  b.battery_charge = soc;
```
stevelimyh:
```suggestion
  b.battery_charge = soc * 100.0;
```
`battery_charge` is documented as percent, but this assigns the RMF state of charge unscaled.

vda5050_core/python/src/rmf_migration.cpp
```diff
@@ -0,0 +1,540 @@
...
+    });
+
+  adapter->on_localize(
+    [callbacks](client::adapter::LocalizationRequest request, auto execution) {
+      auto destination = Destination(
+        request.map_id(), {request.x(), request.y(), request.theta()});
+      auto command = CommandExecution(execution, ActivityIdentifier());
+
+      callbacks.localize()(std::move(destination), std::move(command));
+    });
```
stevelimyh:
```suggestion
  if (callbacks.localize())
  {
    adapter->on_localize(
      [callbacks](client::adapter::LocalizationRequest request, auto execution) {
        auto destination = Destination(
          request.map_id(), {request.x(), request.y(), request.theta()});
        auto command = CommandExecution(execution, ActivityIdentifier());

        callbacks.localize()(std::move(destination), std::move(command));
      });
  }
```

`on_localize` is currently registered unconditionally, but `RobotCallbacks` constructor leaves `localize_` empty. Any `initPosition` then aborts the process. I ran the Python example and published one, and it aborted with `std::bad_function_call`

Guarding the registration fixes this and falls through to the adapter's default transform path, which looks like the intended behaviour when `localize` is not set.

vda5050_core/python/src/rmf_migration.cpp
sauk2: https://github.com/ros-industrial/vda5050_core/pull/71/commits/29c08d5b934d6d0fa3ff7d6fb6cb3ddcf619b910

vda5050_core/src/client/adapter/adapter.cpp
sauk2: https://github.com/ros-industrial/vda5050_core/pull/71/commits/9a0998f5def013e8d381f9abe9d9a5ade9b9a92f

vda5050_core/python/src/rmf_migration.cpp
sauk2: https://github.com/ros-industrial/vda5050_core/pull/71/commits/e32ae5a4c11cbbe9108ecfcfa4e1260a099494e4

vda5050_core/python/src/rmf_migration.cpp
sauk2: https://github.com/ros-industrial/vda5050_core/pull/71/commits/e32ae5a4c11cbbe9108ecfcfa4e1260a099494e4

vda5050_core/src/client/adapter/adapter.cpp
sauk2: https://github.com/ros-industrial/vda5050_core/pull/71/commits/53d0f65ce7f9a70a73d2ba829d0d49e696b1b1a7

vda5050_core/src/client/adapter/adapter.cpp
sauk2: https://github.com/ros-industrial/vda5050_core/pull/71/commits/a4976b7364b033142362c04e713ce21b89dbb463

vda5050_core/src/client/adapter/adapter.cpp
sauk2: https://github.com/ros-industrial/vda5050_core/pull/71/commits/e50dde2c38d44ea2184d3f37e578d38760a64a95

vda5050_core/src/client/adapter/state_manager.cpp
sauk2: Addressed part of this in https://github.com/ros-industrial/vda5050_core/pull/71/commits/877cd4437c8891b0217f149e079f12e3480522c5

For now I have kept the state manager and the publish thread separate till we get around to making strategies that handle this

vda5050_core/src/client/adapter/adapter.cpp
sauk2: I have removed the default localization behaviour. I felt it complicated things by keep the user uninformed about transforms being fixed. Instead I have changed it to fail the `initPosition` action immediately if no localization callback is discovered

vda5050_core/src/client/adapter/adapter.cpp
sauk2: https://github.com/ros-industrial/vda5050_core/pull/71/commits/7fc29225f40a350a24ece635f30dea61c7aed1fb

vda5050_core/src/client/adapter/adapter.cpp
sauk2: Added a check as part of https://github.com/ros-industrial/vda5050_core/pull/71/commits/7fc29225f40a350a24ece635f30dea61c7aed1fb

vda5050_core/src/client/adapter/adapter.cpp
```diff
@@ -0,0 +1,724 @@
...
+          }
+
+          for (const auto& e : result.warnings())
+          {
+            state_manager->add_error(e);
+          }
+
+          request_state_publish();
+          return;
+        }
```
stevelimyh: One more thing I noticed,  I think a new order is dropped after the AGV completes the previous one.

`active->order` does not appear to be cleared when the route finishes, so `has_value()` stays true for the rest of the session. A later order with a different `orderId` then falls through to `is_valid_update()`, which fails on the id mismatch. The errors are added to state, but nothing is logged, so the order appears to disappear.

I reproduced this by running the master example against the adapter:

- `order-0` completed
- the master published `order-1` on the order topic
- the adapter never accepted it — no accept, no reject, no log

A second run with the same `orderId` and incremented `orderUpdateId` was accepted, so this looks specific to the new-order path after completion.

vda5050_core/src/client/adapter/state_manager.cpp
sauk2: https://github.com/ros-industrial/vda5050_core/pull/71/commits/7cadba7fe2c5066b589c7466c5cb46e67ce13fd0

vda5050_core/src/client/adapter/adapter.cpp
sauk2: I made the change as part of https://github.com/ros-industrial/vda5050_core/pull/71/commits/b731d6f0e1c1f4a8de43cb255b26791db40e71e2

Could you test it again?

stevelimyh: Confirmed, this works now. Thanks.

vda5050_core/src/client/adapter/adapter_impl.hpp
sauk2:
```bash
#ifndef header guard has wrong style, please use: CLIENT__ADAPTER__ADAPTER_IMPL_HPP_  [build/header_guard] [5] 
#endif line should be "#endif  // CLIENT__ADAPTER__ADAPTER_IMPL_HPP_"  [build/header_guard] [5]
```

Changing this will fail `cpplint`

stevelimyh (approved):
Thanks for the changes. LGTM

Issue #42, opened 2026-04-22
Remove type identification to remove RTTI dependency


sauk2:
## Summary

The current implementation of the `Provider` and `Context` relies on `std::type_index` and `typeid()` for message dispatching and storage. In UE5, RTTI is disabled by default (bEnableRTTI = false) to optimize performance and reduce binary size.

### Proposed solution

Replace `std::type_index` with a Static Type ID system. This would use an internal counter and template specialization to assign a unique `uint32_t` to every update and event type at compile-time/initialization, bypassing the need for RTTI entirely.

sauk2:
cc @Briancbn @DillonChew98

PR #75, opened 2026-07-02
feat(cpp20): add fixes for c++20 to allow external packages utilize INTERFACE targets

Briancbn:
This PR is primarily for 3rd party `c++20` based applications built on the `vda5050_core` headers.
I encounter this problem when I try to utilize the serialization within the `VDA5050Wrapper` Unreal Plugin Module.
This can temporarily solved in `Unreal` the same way we address the RTTI problem (#42).

Key issues for C++20
- [ambiguous overload error log](https://github.com/Briancbn/vda5050_client/actions/runs/28568655917/job/84701238906)
- [string joining error](https://github.com/Briancbn/vda5050_client/actions/runs/28568655917/job/84701238889)

Key changes made
- use `get_to` for ambiguous type such as `std::vector<T>`
- use `fmt` pattern for string joining

I also added a pipeline to test building for c++20

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,6 @@ on:
     branches:
       - main
       - develop
-      - test/bn/cpp20-error-fix
 
 jobs:
   industrial_ci:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ on:
     branches:
       - main
       - develop
-      - test/bn/cpp20-error
+      - test/bn/cpp20-error-fix
 
 jobs:
   industrial_ci:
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
sauk2, line 17: I think `rolling` is on C++20 right now. do you also want to add `lyrical` and `rolling` to the existing CI

Briancbn: > do you also want to add lyrical and rolling to the existing CI

Yes!

I am more inclined do this separately for the additional ROS2 distro support.
Since we currently [pin our version to `17`](https://github.com/ros-industrial/vda5050_core/blob/develop/vda5050_core/CMakeLists.txt#L5), we still need to enable `20` manually in the same manner in the new distros.

Issue #61, opened 2026-06-02
Create a strongly-typed VDA5050 protocol version


sauk2:
## Summary

The internal configurations currently store the protocol version as a free-form `std::string`. This allows any arbitrary value to be stored, including invalid or unsupported versions. Since the library only supports a defined set of VDA5050 protocol versions, it would be a good idea to introduce a strongly-typed protocol version representation

## Proposed Implementation

Introduce a dedicated protocol version type that acts as the canonical representation of supported VDA5050 versions.

This type should become the single source of truth for:
- Supported protocol versions.
- Conversion between internal representations and external string representations.
- Generation of protocol-specific topic version strings.

Additionally, existing version-conversion logic within the protocol adapter can be transferred into this class once it is implemented.

https://github.com/ros-industrial/vda5050_core/blob/7d92951484d65c4b5bc97c16f78dcb1955d1258d/vda5050_core/include/vda5050_core/execution/protocol_adapter.hpp#L220-L226

leahmarymathew:
hi can i work on this issue ?

PR #76, opened 2026-07-02
refactor(types): migrate protocol version handling to ProtocolVersion (#61)

leahmarymathew:
This adds a proper ProtocolVersion type instead of passing raw version strings around, then migrates ProtocolAdapter, HeaderConfigResource, and master::standard_names::Version to use it. get_topic_version() on ProtocolAdapter goes away since ProtocolVersion::topic_version() replaces it.

One side effect worth flagging: header.version now correctly emits the semantic version (2.0.0) instead of the topic version (v2) - these were previously getting conflated since they came from the same untyped string.

Tests are updated to match. Refs #61.

## Test plan
- [ ] Existing test suite passes with updated assertions
- [ ] CI green on develop


vda5050_core/include/vda5050_core/types/protocol_version.hpp
```diff
@@ -0,0 +1,115 @@
...
+namespace types {
+
+/// \brief Strongly-typed VDA5050 protocol version.
+///
+/// Canonical representation of the VDA5050 protocol versions supported by
+/// this library. Single source of truth for which versions are supported,
+/// conversion to the full semantic version string used in message headers
+/// (e.g. "2.0.0"), and conversion to the MQTT topic version segment (e.g.
+/// "v2").
+class ProtocolVersion
```
sauk2, line 38: I think the current design of this class feels a bit heavy because adding a new version will require changes at multiple places in this code. Can we instead simplify this to an enum class with a bunch of free functions because to me to usage looks similar?

sauk2 (changes requested):
Thanks for the contribution! I have provided my feedback.

Before adding new commits, I would suggest you merge the new changes from `develop` into your branch

vda5050_core/include/vda5050_core/types/protocol_version.hpp
sauk2: I think you can ignore my previous comment. After internal discussion, we are leaning more towards a simpler value-type approach for `ProtocolVersion`.

The idea is to keep `ProtocolVersion` as a small class constructed from `major`, `minor`, `patch` and then define the supported versions as `inline constexpr`.

An example implementation could be like,

```cpp
class ProtocolVersion
{
public:
  constexpr ProtocolVersion(uint8_t major, uint8_t minor, uint8_t patch);
  ...
  ...
  ...
};
```
This class can have functions like `to_string()` and `to_topic_version()`

Then the internal construction will look like,
```cpp
namespace protocol_versions {
  inline constexpr ProtocolVersion V2_0_0{2, 0, 0};
}
```

The usage would be the following,
```cpp
auto version = protocol_versions::V2_0_0;
```

leahmarymathew:
Thank you for the feedback I've made the changes as per the instruction. I'd appreciate it if you could take another look when you have time.

vda5050_core/include/vda5050_core/types/protocol_version.hpp
```diff
@@ -0,0 +1,126 @@
...
+{
+public:
+  constexpr ProtocolVersion(uint8_t major, uint8_t minor, uint8_t patch)
+  : major_(major), minor_(minor), patch_(patch)
+  {
+    // Nothing to do here ...
+  }
```
sauk2: I was trying to imagine the usage here and I realized giving out the constructor publicly invites users to create any version they want. So, I think I will ask you to back-track to a simpler variation of your previous implementation.

You can make the constructor private and the expose a `static const`

```cpp
class ProtocolVersion
{
public:
  static const ProtocolVersion V2_0_0;

  ...
  ...

private:
  constexpr ProtocolVersion(uint8_t major, uint8_t minor, uint8_t patch)
  : major_(major), minor_(minor), patch_(patch)
  {
    // Nothing to do here ...
  }

  uint8_t major_;
  uint8_t minor_;
  uint8_t patch_;
};

inline constexpr ProtocolVersion ProtocolVersion::V2_0_0{2, 0, 0};
```

vda5050_core/include/vda5050_core/types/protocol_version.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+  /// \brief Inequality operator
+  ///
+  /// \param other The other object to compare to
+  ///
+  /// \return is not equal?
+  constexpr bool operator!=(const ProtocolVersion& other) const
+  {
+    return !(this->operator==(other));
+  }
```
sauk2, line 78: Can you also add `<=` and `>=` operators?

vda5050_core/include/vda5050_core/types/protocol_version.hpp
```diff
@@ -0,0 +1,126 @@
...
+private:
+  uint8_t major_;
+  uint8_t minor_;
+  uint8_t patch_;
+};
+
+namespace protocol_versions {
+inline constexpr ProtocolVersion V2_0_0{2, 0, 0};
+inline constexpr std::array<ProtocolVersion, 1> kSupportedVersions{V2_0_0};
+}  // namespace protocol_versions
```
sauk2: You can remove the namespace completely and store the supported versions inside the class itself

vda5050_core/include/vda5050_core/types/protocol_version.hpp
```diff
@@ -0,0 +1,126 @@
...
+inline ProtocolVersion ProtocolVersion::from_string(const std::string& version)
+{
+  for (const auto& supported : supported_versions())
+  {
+    if (supported.to_string() == version) return supported;
+  }
+
+  throw std::invalid_argument(
+    "ProtocolVersion::from_string: unsupported protocol version '" + version +
+    "'");
```
sauk2: I would suggest not to throw an exception and instead return an optional value from this function

```cpp
static std::optional<ProtocolVersion> from_string(const std::string& version);
```

sauk2 (changes requested):
Thanks for the changes! Apologies to ask you to revert some changes to your previous implementation. I think this PR will be ready to be merged once this feedback is addressed :rocket:

PR #78, opened 2026-07-03
refactor(layout): use nested namespaces in headers for C++14 compatibility

stevelimyh:
## Summary

Fixes C++14 compatibility in the layout headers.

Collapsed namespace syntax (`namespace vda5050_core::layout`) is C++17-only, so this PR converts the 14 affected headers to expanded nested namespaces.

## What's included

- Updated 13 `layout/` headers + `layout_serialization.hpp`
- Header-only change
- No `.cpp` changes


Briancbn (approved):
Lgtm!

PR #80, opened 2026-07-22
fix(client): transform node positions if a transformation is available

sauk2:
## Summary

There is a bug where the node positions do not get transformed into AGV frame before being sent to user's navigation callback. This PR implements a check for the presence of transformations before sending a navigation request.


stevelimyh (approved):
Fix looks right, and the round trip is consistent now: `on_navigate` receives AGV frame and `set_position` takes AGV frame back.

PR #82, opened 2026-07-22
fix(master): order rejection callback

stevelimyh:
## Summary
  `assign_order` returns `ASSIGNED` once the order is queued. The rest of the
  validation runs afterwards on the AGV's outbound worker, and if it rejects
  there the errors are logged and dropped: no callback, no state change, nothing
  to query. The order never reaches the AGV, so nothing comes back in State
  either.

  This adds `on_order_rejected(agv_id, order_id, errors)`, pairing with
  `on_order_complete` so both endings of an assigned order arrive the same way.

### DIFF
--- a/vda5050_core/examples/master/master_example.cpp
+++ b/vda5050_core/examples/master/master_example.cpp
@@ -22,6 +22,7 @@
 #include <memory>
 #include <string>
 #include <thread>
+#include <vector>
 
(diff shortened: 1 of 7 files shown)


sauk2 (approved):
Thanks for the fix! I couldn't find any issue. Tested locally and there is no change to the existing behaviour

PR #84, opened 2026-07-24
docs(api): add doxygen file for API documentation generation

Briancbn:
The XML config is needed for https://github.com/ros-industrial/rmf2-docs/pull/1

Install Doxygen
```bash
sudo apt install doxygen graphviz
```

Run 
```bash
doxygen
```

Open Firefox
```bash
firefox docs/api/html/index.html
```


docs/api/.gitignore
sauk2, line 1: Is there a need to have a separate `.gitignore`? Can this be put into the root file?

Doxyfile
```diff
@@ -0,0 +1,31 @@
...
+PROJECT_NAME           = "VDA5050 Core"
+PROJECT_NUMBER         =
+PROJECT_BRIEF          =
+OUTPUT_DIRECTORY       = ./docs/api
+EXTRACT_ALL            = YES
+EXTRACT_PRIVATE        = NO
+EXTRACT_PACKAGE        = NO
+EXTRACT_STATIC         = YES
+INPUT                  = ./README.md \
+                         ./vda5050_core/include
```
sauk2, line 13: Can this be formatted to be on the same line?

```suggestion
INPUT                  = ./README.md ./vda5050_core/include
```

sauk2 (approved):

Doxyfile
Briancbn: I separated them, so that it's easier for us to add new packages without looking up how doxyfile syntax works

docs/api/.gitignore
Briancbn: oh, i remember that for some tools if we don't have the folder, we need to add additional instructions for `mkdir`.

It seems that this is not needed for `doxygen`. I will combine it.

PR #86, opened 2026-07-24
fix(client): pass action execution param as json

sauk2:
## Summary

This PR adds a few fixes to make the migration follow Open-RMF fleet adapter pattern more closely. 
- Fixes the arguments for the action executor callback
- Changes the activity identifier to make construction and comparison simpler
- Provides bindings for transformation classes
- Exposes the transformation related state manager functions
- Accepts `startPause` as way to invoke the `stop` callback


stevelimyh (approved):
Thanks for the changes. I didn’t find any issues.

PR #88, opened 2026-08-13
feat(pybind): refactor pybind code for better compatibility with modern pyproject

Briancbn:
This PR combine aspects of #70 and #85 and serve as a prerequisite before those features.

Key changes
- Remove reliance on `ament_cmake`. Only the following 2 aspect rely on `ament_cmake` due to their limited impact on development environment
  - python packaging for `pip show` ([ref](https://github.com/Briancbn/vda5050_client/blob/a8bf8a91418b7844b9649e1ef0a65df2d2d77fa2/vda5050_core/python/CMakeLists.txt#L55-L59))
  - ament linters
- Add `pyproject.toml` and `scikit-build-core` compatibility following [scikit_build_example](https://github.com/pybind/scikit_build_example), while still have compatibility within the `colcon` environment.
- Restructuring of the pybind `.hpp` and `.cpp`.
  - Separate bindings for different **python submodules** to separate files. This also prepares for bindings for **master** and more
  - Change the target name to `_core`
  - Change python package name to `vda5050_core` and add import files to flatten the namespaces
  - Add stubs for autocompletion and documentation generation, (generation instructions included in `developer.md`)
- Add CI for ruff formatter and build for different python version.

Now python import is changed to 
```python
from vda5050_core.rmf_migration import Adapter

Adapter.make(...)
```

## How to use

Install build dependencies
```sh
sudo apt-get install \
  libpaho-mqtt-dev libpaho-mqttpp-dev \
  libfmt-dev \
  nlohmann-json3-dev \
  pybind11-json-dev
```

### COLCON (no ROS)

simply
```sh
colcon build
```
you should see a warning statement here

test import
```sh
source install/setup.bash
python3 -c "from vda5050_core.rmf_migration import Adapter; print('Import successfully')"
```

### COLCON (with ROS)

simply
```sh
source /opt/ros/<ros-distro>/setup.bash
colcon build
```
no warning message

test import
```sh
source install/setup.bash
python3 -c "from vda5050_core.rmf_migration import Adapter; print('Import successfully')"
```

### pip

virtual environment high recommended
```sh
python3 -m venv vda5050-venv
. vda5050-venv/bin/activate
```

Setup using pip
```sh
cd <repo-root>/vda5050_core
pip install .[test]
```

run tests
```sh
pytest .
```

### UV

install (editable)
```sh
cd <repo-root>/vda5050_core
uv sync --group test
```

run test
```sh
uv run pytest
```

build `.whl`, `dist` folder
```sh
uv build
```
[...]

PR #89, opened 2026-08-14
feat(pybind): add cibuildwheel support

Briancbn:
Depend on #88 

~~Meanwhile you can check the diff from below~~
~~https://github.com/Briancbn/vda5050_client/compare/feat/bn/pybind-packaging...Briancbn:vda5050_client:feat/bn/add-cibw-support~~

The pipeline takes heavy inspiration from [numpy](https://github.com/numpy/numpy/blob/main/.github/workflows/wheels.yml)

- I can clean up the comments later, but this is ready for review.
- Let me know if I should separate the MacOS support into a different PR.

I only select 1 job for each architecture
We will probably make another repo for releases pipeline similar to [numpy/numpy-release](https://github.com/numpy/numpy-release) for all the architectures. 

> side note, they also release nightly version into conda-forge, which we can also consider.

How to run `cibuildwheel` locally

```sh
cd <repo-root>/vda5050_core
uvx cibuildwheel
```

After completion, you can find the `.whl` files under `wheelhouse` folder

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -85,6 +85,9 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: "Install UV"
+        uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
+
       - name: Install package and test dependencies
         run: uv sync --group dev
 
@@ -100,7 +103,7 @@ jobs:
 
   wheels:
     name: Build wheels ${{ matrix.python }}-${{ matrix.buildplat[1] }}
-    needs: [lint, test]
+    needs: [lint, test, stub-test]
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -107,12 +107,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ["cp312"]
+        python: ["cp310", "cp312"]
         buildplat:
           - [ubuntu-latest, manylinux_x86_64]
-          - [ubuntu-latest, musllinux_x86_64]
           - [ubuntu-24.04-arm, manylinux_aarch64]
-          - [ubuntu-24.04-arm, musllinux_aarch64]
           - [macos-15-intel, macosx_x86_64]
           - [macos-15, macosx_arm64]
     runs-on: ${{ matrix.buildplat[0] }}
@@ -127,11 +125,7 @@ jobs:
       - name: Setup macOS
         if: startsWith(matrix.buildplat[1], 'macosx_')
         run: |
-          if [[ "${{ matrix.buildplat[1] }}" == "macosx_arm64" ]]; then
-            CIBW="MACOSX_DEPLOYMENT_TARGET=14.0"
-          else
-            CIBW="MACOSX_DEPLOYMENT_TARGET=15.0"
-          fi
+          CIBW="MACOSX_DEPLOYMENT_TARGET=15.0"
           echo "CIBW_ENVIRONMENT_MACOS=$CIBW" >> "$GITHUB_ENV"
 
       - name: Build wheels
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -114,7 +114,7 @@ jobs:
           - [ubuntu-24.04-arm, manylinux_aarch64]
           - [ubuntu-24.04-arm, musllinux_aarch64]
           - [macos-15-intel, macosx_x86_64]
-          - [macos-14, macosx_arm64]
+          - [macos-15, macosx_arm64]
     runs-on: ${{ matrix.buildplat[0] }}
     steps:
       - uses: actions/checkout@v6
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -64,9 +64,6 @@ jobs:
   stub-test:
     name: Check Stale Stubs
     runs-on: ubuntu-latest
-    defaults:
-      run:
-        working-directory: vda5050_core
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -8,13 +8,14 @@ on:
       - main
       - develop
 
+defaults:
+  run:
+    working-directory: vda5050_core
+
 jobs:
   lint:
     name: Ruff Lint
     runs-on: ubuntu-latest
-    defaults:
-      run:
-        working-directory: vda5050_core
     steps:
       - uses: actions/checkout@v6
 
@@ -30,13 +31,10 @@ jobs:
   test:
     name: UV Test (Python ${{ matrix.python-version }})
     runs-on: ubuntu-latest
-    defaults:
-      run:
-        working-directory: vda5050_core
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12"]
+        python-version: ["3.12"]
     steps:
       - uses: actions/checkout@v6
 
@@ -90,9 +88,6 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: "Install UV"
-        uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
-
       - name: Install package and test dependencies
         run: uv sync --group dev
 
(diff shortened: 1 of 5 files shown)


Briancbn:
This PR is ready for review. @sauk2 

A few more points here on the `install-native-deps.sh`

### manylinux

The `manylinux` image is `almalinux 8.10` (RHEL like) that uses the `dnf` and `yum` (interchangeable) for installing dependencies.
Unfortunately the `pybind11-devel` and `pybind11-json-devel` seems to be broken
- `pybind11-devel` ([rosdistro](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L9217)), doesn't exist, instead only `python3.x-pybind11-devel` is available for each python version.
- `pybind11-json-devel` exists, but depend on `pybind11-devel` that doesn't exist. So installation is broken

More on this https://github.com/ros/rosdistro/pull/42740

You can check this by running in docker 
```shell
docker run -it --rm quay.io/pypa/manylinux_2_28_x86_64
```

### musllinux

The `musllinux` image is `Alpine 3.22.5` that uses `apk` for installing dependencies. 
Registry: <https://pkgs.alpinelinux.org/packages>
- `pybind11`, `pybind11-json` and `paho-mqtt` doesn't exist here

You can check this by running in docker
```shell
docker run -it --rm  quay.io/pypa/musllinux_1_2_x86_64
```

### MacOS

It seems that `14.0` is the oldest we can support for ARM based and `15.0` is the oldest we can support for Intel based MacOS. These are primarily due to `openssl`. The current version in use is `20+` (26), so I think we are ok.

More on this: 
- <https://cibuildwheel.pypa.io/en/stable/cpp_standards/#macos-and-deployment-target-versions>

@GameTL can help check also.

### Windows

Not supported so far, a lot of things are different. I will address this in #70

.github/workflows/python.yml
```diff
@@ -105,3 +100,47 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        python: ["cp312"]
+        buildplat:
+          - [ubuntu-latest, manylinux_x86_64]
+          - [ubuntu-latest, musllinux_x86_64]
+          - [ubuntu-24.04-arm, manylinux_aarch64]
+          - [ubuntu-24.04-arm, musllinux_aarch64]
+          - [macos-15-intel, macosx_x86_64]
+          - [macos-14, macosx_arm64]
```
sauk2:
```suggestion
          - [macos-15, macosx_arm64]
```
What do you think about instead building on `macos-15` while setting `MACOSX_DEPLOYMENT_TARGET=14.0`? I think Sonoma goes EOL sometime this year

sauk2:
> ### musllinux
> The `musllinux` image is `Alpine 3.22.5` that uses `apk` for installing dependencies. Registry: https://pkgs.alpinelinux.org/packages
> 
> * `pybind11`, `pybind11-json` and `paho-mqtt` doesn't exist here
> 
> You can check this by running in docker
> 
> ```shell
> docker run -it --rm  quay.io/pypa/musllinux_1_2_x86_64
> ```

Do we have a strong requirement to provide `musllinux` wheels so early on? If not, then would it make sense to land `manylinux` and `macOS` support fisrt and add `musllinux` support separately later?

.github/workflows/python.yml
```diff
@@ -105,3 +100,47 @@ jobs:
...
           echo "Stubs are stale!!"
           exit 1
+
+  wheels:
+    name: Build wheels ${{ matrix.python }}-${{ matrix.buildplat[1] }}
+    needs: [lint, test, stub-test]
+    strategy:
+      fail-fast: false
+      matrix:
+        python: ["cp312"]
```
sauk2: In the `pyproject.toml` we support python versions from `3.10` onwards but we are only making a wheel installable on `3.12`. Is this intentional for the initial wheel release?

vda5050_core/scripts/install-native-deps.sh
```diff
@@ -0,0 +1,167 @@
+#!/bin/bash
+# Install native build deps for vda5050 (fmt, nlohmann_json, Paho MQTT C++).
+set -euo pipefail
+
+PAHO_CPP_TAG="${PAHO_CPP_TAG:-v1.6.0}"
+FMT_TAG="${FMT_TAG:-11.2.0}"
+NLOHMANN_JSON_TAG="${NLOHMANN_JSON_TAG:-v3.11.3}"
+PYBIND11_TAG="${PYBIND11_TAG:-v2.13.6}"
+PYBIND11_JSON_TAG="${PYBIND11_JSON_TAG:-0.2.13}"
```
sauk2: Can we pin it to commit hashes instead of tags since tags are mutable?

vda5050_core/scripts/install-native-deps.sh
```diff
@@ -0,0 +1,167 @@
...
+    echo "Unsupported Linux package manager" >&2
+    exit 1
+  fi
+}
+
+install_json_from_source() {
+  local work
+  work="$(mktemp -d)"
+
+  git clone --depth 1 --branch "$NLOHMANN_JSON_TAG" \
```
sauk2: I see that tags are being used when installing from source but not when installing through package managers. Are we okay with having floating versions for Homebrew and APT package manager?

sauk2 (changes requested):
Provided my initial comments

.github/workflows/python.yml
```diff
@@ -105,3 +100,47 @@ jobs:
...
+          else
+            CIBW="MACOSX_DEPLOYMENT_TARGET=15.0"
+          fi
+          echo "CIBW_ENVIRONMENT_MACOS=$CIBW" >> "$GITHUB_ENV"
+
+      - name: Build wheels
+        run: >
+          uvx cibuildwheel
+          --only "${{ matrix.python }}-${{ matrix.buildplat[1] }}"
+          --output-dir ../wheelhouse
```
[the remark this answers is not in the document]
Briancbn: The `test-command` defined here runs in the `cibuildwheel` job. You can test it out by intention modifying the test to fail.

https://github.com/Briancbn/vda5050_client/blob/c45373d988bd81c35a98043490af2312e82f32ab/vda5050_core/pyproject.toml#L68

.github/workflows/python.yml
Briancbn: I agree. @GameTL what do you think?

.github/workflows/python.yml
Briancbn: Yes, I can add `cp310` in too and remove the `musllinux`. We will build for all the platform that we support, but we can do that in a separate `vda5050_core-release` repo.

vda5050_core/scripts/install-native-deps.sh
Briancbn: Good questions. Should I just install **all** dependencies from source? `apt` removes old dependencies when newer ones are introduced for certain packages.

sauk2: Yes, I think we can try that. That way all platforms will have the same versions of dependencies

.github/workflows/python.yml
GameTL: yes I think we can bump to 15 is good as people don't like macos 26 more than usual.

GameTL (approved):

GameTL (changes requested):
accidental approve. after the change from tag to hash commit for the dependancy for the CIs

vda5050_core/scripts/install-native-deps.sh
GameTL: https://github.com/eclipse-paho/paho.mqtt.cpp/releases/tag/v1.6.0 
like this commit? i see both mac and linux build ci's are using this 

do we need to do checks for the paho mqtt c lib which the cpp version depends on?

Briancbn: All dependencies changed to install from source https://github.com/ros-industrial/vda5050_core/pull/89/changes/472c0cfeb17747729820a7a692f9b90d8d88a835

vda5050_core/scripts/install-native-deps.sh
Briancbn: Tagged to commit in https://github.com/ros-industrial/vda5050_core/commit/472c0cfeb17747729820a7a692f9b90d8d88a835


Add in bash functions for shallow clone in https://github.com/ros-industrial/vda5050_core/pull/89/changes/d1b48da302c9c382a3edfb25daf647b3432b94bf.

some additional shellcheck fixes in https://github.com/ros-industrial/vda5050_core/pull/89/changes/bec497b3074f9093223bf04864209d1f52dbb27c
Let me know if the bash function is ok.

.github/workflows/python.yml
Briancbn: addressed in https://github.com/ros-industrial/vda5050_core/pull/89/changes/5138bfc6a36f4f05451f8b01df3fe179646686a1

sauk2 (approved):
I am okay to merge :rocket:

Thanks for the great work!

GameTL (approved):
beautiful

PR #91, opened 2026-08-21
fix(client): fix execution lifecycle handling for navigation and actions

sauk2:
## Summary

`CommandExecution` under `rmf_migration` holds the underlying `Execution` as a base-class pointer. Because the completion methods were implemented differently across the base and derived navigation and action execution classes, calling them from Python would result in the base class function being invoked due to downcasting behavior.

This caused executions that were successfully completed by the Python callback to remain active or be reported incorrectly as failed.

### Solution

Changed the execution API so that completion is handled through the appropriate virtual interface on the base `Execution` class, while the derived `ActionExecution`/`OrderExecution` classes perform their VDA5050 status callback before marking the underlying execution as complete.

### DIFF
--- a/vda5050_core/include/vda5050_core/client/adapter/action_execution.hpp
+++ b/vda5050_core/include/vda5050_core/client/adapter/action_execution.hpp
@@ -46,9 +46,11 @@ public:
 
   void paused(std::optional<std::string> result_description = std::nullopt);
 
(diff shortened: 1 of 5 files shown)


Briancbn (approved):
Thanks for addressing this quickly.

Apart from the extra line for DCO testing, lgtm

stevelimyh (approved):
LGTM!

